PR #222, opened 2026-04-07
feat(crawler): add web crawler for site content ingestion

### LINKED ISSUE #132: feat(crawler): Add web crawler for automatic site content ingestion
## Context

Currently, professors build their chatbot's knowledge base by **manually uploading files** (PDFs, documents, etc.). However, many professors want to deploy the chatbot on a website (course site, department wiki, documentation portal) and have it **automatically understand the site's content**.

This feature adds a web crawler that can:
1. **Auto-discover** pages from a root URL (crawl the site)
2. **Allow manual URL input** for specific pages to include/exclude
3. **Extract, chunk, and embed** page content into the existing RAG pipeline

This transforms the chatbot from a "file-based Q&A bot" into a "site-aware assistant" that can answer questions about any content on the site where it's deployed.

## Relevant Code

| File | Purpose |
|------|---------|
| `packages/ai/src/rag-service.ts` | RAG pipeline — chunking (`chunkText`) and embedding generation; crawler content feeds into this |
| `packages/ai/src/openrouter-client.ts` | Embedding generation via OpenAI `text-embedding-3-small` |
| `packages/db/src/schema.ts` | Database schema — will need new tables for crawled pages |
| `apps/web/src/lib/file-processor.ts` | Existing file processing pattern — crawler processing follows similar async job pattern |
| `apps/web/src/lib/qstash.ts` | QStash job publishing — use for async crawl jobs |
| `apps/web/src/app/api/jobs/process-file/route.ts` | Example of a QStash job handler — create similar for crawl jobs |
| `apps/web/src/server/routers/chatbot.ts` | Chatbot settings — add crawler configuration here |

### Existing async processing pattern:

```
1. User triggers action (tRPC endpoint)
2. Server publishes job to QStash
3. QStash calls /api/jobs/<job-name> endpoint
4. Job handler processes asynchronously
5. Status/progress stored in DB (polled by client)
```

The crawler should follow this exact same pattern.

### 1. Database: New tables for crawled content

Add to `packages/db/src/schema.ts`:

- **`crawlSources`** table: Tracks root URLs configured for a chatbot
  - `id`, `chatbotId` (FK), `rootUrl`, `status` (pending/crawling/completed/failed), `crawlDepth`, `maxPages`, `includePatterns` (JSONB array of URL patterns to include), `excludePatterns` (JSONB array), `lastCrawledAt`, `metadata` (JSONB: pageCount, errors), `createdAt`, `updatedAt`

- **`crawledPages`** table: Individual pages discovered and processed
  - `id`, `crawlSourceId` (FK), `url`, `title`, `content` (extracted text), `contentHash` (for change detection on re-crawl), `status` (pending/processing/completed/failed), `metadata` (JSONB: statusCode, contentType, depth, wordCount), `createdAt`, `updatedAt`

- Crawled page content should feed into the existing `fileChunks` table for embedding storage (or a new `pageChunks` table with the same structure)

### 2. Backend: Crawler engine

Create a crawl processing module:

- **URL discovery**: Given a root URL, discover linked pages on the same domain
  - Respect `robots.txt` rules
  - Follow links up to configurable depth (default: 3)
  - Stay within the same domain (no external links)
  - Configurable max pages limit (default: 100)
  - Support include/exclude URL patterns (e.g., include `/docs/*`, exclude `/admin/*`)

- **Content extraction**: For each discovered page
  - Fetch HTML content
  - Extract meaningful text (strip nav, header, footer, scripts, styles)
  - Extract page title and metadata
  - Use a library like `cheerio` for HTML parsing or `mozilla/readability` for article extraction

- **RAG integration**: Feed extracted text into the existing pipeline
  - Chunk text using `chunkText()` from `rag-service.ts`
  - Generate embeddings using the existing embedding function
  - Store in chunks table, associated with the chatbot

- **Change detection**: On re-crawl, compare content hashes to only re-process changed pages

### 3. Backend: tRPC API

Add new procedures (new router or extend chatbot router):

- `addCrawlSource`: Add a root URL to crawl for a chatbot
  - Input: `{ chatbotId, rootUrl, crawlDepth?, maxPages?, includePatterns?, excludePatterns? }`
  - Validates URL format, verifies chatbot ownership
  - Creates `crawlSources` record and triggers async crawl job

- `removeCrawlSource`: Remove a crawl source and its pages/chunks
  - Cascading delete of crawled pages and their chunks

- `getCrawlSources`: List all crawl sources for a chatbot with status

- `getCrawledPages`: Paginated list of crawled pages for a source with status

- `recrawl`: Trigger a re-crawl of an existing source (updates changed pages only)

- `addManualUrl`: Add a specific URL to crawl (not auto-discovered)
  - Input: `{ chatbotId, url }`
  - Crawls just that single page

### 4. Async processing

- Create `/api/jobs/crawl-site/route.ts` — QStash job handler for crawling
- Processing stages with progress reporting:
  1. **Discovering** (0-20%): Crawl site and discover pages
  2. **Fetching** (20-50%): Download page content
  3. **Extracting** (50-70%): Parse HTML and extract text
  4. **Embedding** (70-95%): Chunk and generate embeddings
  5. **Storing** (95-100%): Save to database
- Handle errors per-page (don't fail entire crawl if one page errors)

### 5. Frontend: Crawler management UI

Add a **"Web Sources"** tab to the chatbot detail page:

- **Add source form**: Input for root URL, optional depth/max pages settings
- **Source list**: Shows all crawl sources with status, page count, last crawled date
- **Source detail**: Expandable list of crawled pages with status indicators
- **Manual URL input**: Simple form to add individual URLs
- **Actions**: Recrawl, remove source, add/edit include/exclude patterns
- **Progress indicator**: Show crawl progress while running

## Acceptance Criteria

- [ ] Professors can add a root URL and trigger automatic crawling
- [ ] Crawler discovers pages within the same domain up to configured depth
- [ ] `robots.txt` is respected
- [ ] Include/exclude URL patterns work correctly
- [ ] Page content is extracted cleanly (no nav/header/footer/script noise)
- [ ] Extracted content is chunked, embedded, and stored in the RAG pipeline
- [ ] Manual single-URL crawling works
- [ ] Re-crawl only reprocesses changed pages (content hash comparison)
- [ ] Crawl progress is visible in the UI
- [ ] Professors can remove crawl sources (cascading cleanup)
- [ ] Chatbot ownership is verified on all endpoints
- [ ] Error handling: individual page failures don't block the entire crawl
- [ ] UI is responsive and follows existing patterns (Shadcn UI + Tailwind)
- [ ] Database schema changes include proper migrations
- [ ] `npm run check-types` and `npm run lint` pass

## Technical Notes

- **Crawling library**: Consider [`crawlee`](https://crawlee.dev/) (full-featured Node.js crawler), or build lighter with `fetch` + `cheerio` + a URL queue. For this use case, a custom lightweight solution with `cheerio` + `@mozilla/readability` is likely sufficient and avoids heavy dependencies.
- **Content extraction**: [`@mozilla/readability`](https://github.com/mozilla/readability) is excellent for extracting article content from HTML. Combine with `cheerio` for cases where Readability doesn't apply (e.g., documentation pages with sidebars).
- **Rate limiting**: Add a delay between page fetches (e.g., 1-2 seconds) to be a good citizen. Make this configurable.
- **robots.txt parsing**: Use [`robots-parser`](https://www.npmjs.com/package/robots-parser) to check if URLs are crawlable.
- **URL normalization**: Normalize URLs (remove fragments, trailing slashes, query params for dedup) to avoid crawling the same page twice.
- **Scope control**: Be strict about staying within the same domain. Don't follow external links, CDN links, or social media links.
- **Storage**: Crawled content could reuse the `fileChunks` table by adding a nullable `crawledPageId` FK, or use a separate `pageChunks` table. Either approach works — consistency with the existing pattern is preferred.
- **QStash timeout**: Crawling many pages can take a long time. Consider splitting into multiple QStash jobs (one for discovery, then one per page for processing) to avoid timeout issues.

## Contributing Guidelines

- Follow the coding standards in `AGENTS.md` (strict TypeScript, no `any`, kebab-case files, etc.)
- PR title must use conventional commit format: `feat(crawler): add web crawler for site content ingestion`
- This is a large feature — consider breaking your PR into logical commits (schema, crawler engine, API endpoints, UI)
- Run `npm run lint && npm run check-types` before submitting
- Run `npm run db:generate` if schema changes are made
- Link this issue in your PR description

gum1x:
## Summary
- Adds a web crawler that auto-discovers and indexes pages from a root URL into the chatbot's RAG knowledge base
- Professors can configure crawl depth, max pages, include/exclude URL patterns, manually add individual URLs, re-crawl for updates, and export crawled content as JSON
- Crawler respects robots.txt, enforces HTML-only fetching, scopes to same domain, and blocks private IPs

## Why
Professors deploying chatbots on course sites or documentation portals needed a way to automatically ingest web content without manual file uploads. Resolves the assigned issue.

## Changes
- **Schema**: `crawl_sources` and `crawled_pages` tables + migration (`0006_milky_raza.sql`)
- **Crawler engine** (`packages/ai/src/crawler.ts`): BFS discovery, HTML-only fetching via `text/html` content-type check, Mozilla Readability extraction, content hashing for change detection on re-crawl
- **API**: tRPC `crawlerRouter` with 7 procedures + two QStash job handlers (`/api/jobs/crawl-discover`, `/api/jobs/crawl-process-page`) split to avoid timeout issues on large crawls
- **UI**: `WebSourcesTab` on the chatbot detail page with live progress polling (3s interval), per-page status indicators, recrawl/remove/export JSON actions

## Library choices
- `cheerio` — link extraction and DOM traversal (lightweight, no headless browser needed)
- `@mozilla/readability` — article content extraction (same algorithm as Firefox Reader View, strips nav/header/footer noise cleanly)
- `robots-parser` — robots.txt compliance (stable spec, infrequent updates expected)

## Test Plan
- [ ] Add a root URL and verify pages are discovered and embedded
- [ ] Confirm non-HTML URLs (PDFs, images) are skipped
- [ ] Confirm robots.txt-blocked pages are marked "blocked"
- [ ] Trigger a re-crawl and verify unchanged pages are skipped
- [ ] Remove a source and verify cascading cleanup of pages + embeddings
- [ ] Export JSON and verify content matches crawled pages
- [ ] `npm run check-types` and `npm run lint` both pass


packages/db/drizzle/0006_milky_raza.sql
```diff
@@ -0,0 +1,34 @@
...
+	"depth" integer DEFAULT 0 NOT NULL,
+	"status" "crawled_page_status" DEFAULT 'pending' NOT NULL,
+	"metadata" jsonb DEFAULT '{}'::jsonb,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+ALTER TABLE "crawl_sources" ADD CONSTRAINT "crawl_sources_chatbot_id_chatbots_id_fk" FOREIGN KEY ("chatbot_id") REFERENCES "public"."chatbots"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "crawled_pages" ADD CONSTRAINT "crawled_pages_crawl_source_id_crawl_sources_id_fk" FOREIGN KEY ("crawl_source_id") REFERENCES "public"."crawl_sources"("id") ON DELETE cascade ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "crawled_pages" ADD CONSTRAINT "crawled_pages_user_file_id_user_files_id_fk" FOREIGN KEY ("user_file_id") REFERENCES "public"."user_files"("id") ON DELETE cascade ON UPDATE no action;
```
akhileshrangani4: The new tables are missing indexes and this is going to hurt pretty quickly.

`crawled_pages.crawl_source_id` is used in basically every query in this feature (getCrawledPages, all the count queries, finalizeCrawlSource, exportJson, removeCrawlSource). Without an index, all of those are sequential scans. Same for `crawl_sources.chatbot_id`.

Suggested additions:
```sql
CREATE INDEX idx_crawl_sources_chatbot_id ON crawl_sources(chatbot_id);
CREATE INDEX idx_crawled_pages_source_status ON crawled_pages(crawl_source_id, status);
```

Also worth adding a unique constraint on `(crawl_source_id, url)` for `crawled_pages` to prevent duplicate page rows from concurrent recrawls. The code does a check-then-insert but that's not safe under concurrency.

apps/web/src/server/routers/crawler/procedures/recrawl.ts
```diff
@@ -0,0 +1,89 @@
...
+      )
+      .limit(1);
+
+    if (!chatbot) {
+      throw new TRPCError({
+        code: "NOT_FOUND",
+        message: "Chatbot not found",
+      });
+    }
+
```
akhileshrangani4, line 14: This resets the source to `pending` without checking if a crawl is already running. The UI disables the button, but there's no server-side guard.

If old QStash page-processing jobs from a previous crawl are still in flight when a recrawl kicks off, you get both old and new jobs processing the same pages at the same time. That can lead to duplicate embeddings, premature finalization, or conflicting status updates.

```typescript
if (["pending", "discovering", "crawling"].includes(source.status)) {
  throw new TRPCError({ code: "CONFLICT", message: "A crawl is already in progress" });
}
```

apps/web/src/lib/crawl-processor.ts
```diff
@@ -0,0 +1,386 @@
...
+        },
+        updatedAt: new Date(),
+      })
+      .where(eq(crawledPages.id, crawledPageId));
+  }
+}
+
+export async function finalizeCrawlSource(
+  crawlSourceId: string,
+): Promise<void> {
```
akhileshrangani4, line 502: `finalizeCrawlSource` is called after every single page finishes processing (from the `crawl-process-page` route handler). Each call loads ALL pages for the source to check if any are still pending/processing.

For 500 pages, that's 500 calls each scanning 500 rows = 250k row reads total, just for finalization checks. Classic O(n^2).

Swap this to a count query instead of loading all rows:
```typescript
const [result] = await db
  .select({ count: sql<number>`count(*)` })
  .from(crawledPages)
  .where(
    and(
      eq(crawledPages.crawlSourceId, crawlSourceId),
      inArray(crawledPages.status, ['pending', 'processing']),
    )
  );
if (Number(result?.count ?? 0) > 0) return;
```

apps/web/src/lib/crawl-processor.ts
```diff
@@ -0,0 +1,386 @@
...
+        try {
+          await processCrawlPage({ crawledPageId: page.id });
+        } catch (error) {
+          logError(error, "Inline crawl page processing failed", {
+            crawledPageId: page.id,
+          });
+        }
+      } else {
+        await publishQStashJob({
+          url: `${env.NEXT_PUBLIC_APP_URL}/api/jobs/crawl-process-page`,
```
akhileshrangani4: QStash jobs are published one at a time in a sequential loop. For 500 pages, that's 500 sequential HTTP requests to QStash which could take 30-60 seconds.

Consider batching these with something like:
```typescript
const BATCH = 20;
for (let i = 0; i < pageRecords.length; i += BATCH) {
  const batch = pageRecords.slice(i, i + BATCH);
  await Promise.all(batch.map(page =>
    publishQStashJob({
      url: `${env.NEXT_PUBLIC_APP_URL}/api/jobs/crawl-process-page`,
      body: { crawledPageId: page.id },
    })
  ));
}
```

apps/web/src/server/routers/crawler/validation.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from "zod";
+
+export const crawlSourceInput = z.object({
+  chatbotId: z.string().uuid(),
+  rootUrl: z.string().url(),
+  crawlDepth: z.number().int().min(1).max(5).default(3),
+  maxPages: z.number().int().min(1).max(500).default(100),
+  includePatterns: z.array(z.string()).default([]),
```
akhileshrangani4: The include/exclude pattern strings have no length or content constraints. These get turned into regexes in `matchGlob()` via `pattern.replace(/\*/g, ".*")`, which means a user can craft patterns that cause catastrophic backtracking (ReDoS).

Something like `z.string().max(100).regex(/^[a-zA-Z0-9\-_\/.* ]+$/)` would help, plus limiting the array size:
```typescript
includePatterns: z.array(z.string().max(100)).max(10).default([]),
excludePatterns: z.array(z.string().max(100)).max(10).default([]),
```

Or better yet, use a proper glob library like `picomatch` instead of hand-rolling regex conversion.

packages/ai/src/crawler.ts
```diff
@@ -0,0 +1,457 @@
...
+      )
+    ) {
+      continue;
+    }
+
+    discovered.push(entry);
+
+    if (entry.depth >= options.maxDepth) continue;
+
+    await delay(options.delayMs ?? DEFAULT_DELAY_MS);
```
akhileshrangani4, line 501: Each page in the BFS gets a 1.5s delay before fetching. With `maxPages=500`, discovery alone needs ~750 seconds (12.5 minutes) of delay time. But `processCrawlDiscovery` has a 5 minute timeout, and QStash/Vercel serverless functions have their own time limits too.

So the practical max is around 200 pages, not 500. The feature doesn't work as advertised at higher page counts.

A few options: reduce the default delay, make it adaptive (respect `Crawl-Delay` from robots.txt), or restructure discovery to persist found URLs in batches and dispatch further discovery as separate jobs.

apps/web/src/server/routers/crawler/procedures/export-json.ts
```diff
@@ -0,0 +1,84 @@
...
+      });
+    }
+
+    const completedPages = await ctx.db
+      .select()
+      .from(crawledPages)
+      .where(
+        and(
+          eq(crawledPages.crawlSourceId, input.crawlSourceId),
+          eq(crawledPages.status, "completed"),
```
akhileshrangani4, line 28: This loads ALL completed pages and then for each page fires a separate query to load ALL file chunks. Two issues:

1. N+1 query pattern (1 + N queries where N = completed pages)
2. Unbounded memory: 500 pages with 20+ chunks each could produce a JSON response well over 50MB, all held in memory. On Vercel serverless that's going to OOM or timeout.

Consider adding pagination, using a single join query, or generating the export async with a download link.

apps/web/src/server/routers/crawler/procedures/add-crawl-source.ts
```diff
@@ -0,0 +1,128 @@
...
+import { TRPCError } from "@trpc/server";
+import { chatbots, crawlSources } from "@teachanything/db/schema";
+import { isUrlSafe } from "@teachanything/ai/crawler";
+import { env } from "@/lib/env";
+import { publishQStashJob } from "@/lib/qstash";
+import { logInfo, logError } from "@/lib/logger";
+import { crawlSourceInput } from "../validation";
+
+export const addCrawlSourceProcedure = protectedProcedure
+  .input(crawlSourceInput)
```
akhileshrangani4, line 11: None of the crawler mutations have rate limiting. The rest of the codebase uses Upstash rate limiters from `lib/rate-limit.ts` (e.g. `chatbotCreationRateLimit`). Each crawl source can trigger up to 500 page fetches plus 500 embedding API calls, so this is expensive to abuse.

Suggested limits:
- `addCrawlSource`: ~5 per hour per user
- `addManualUrl`: ~20 per hour per user
- `recrawl`: ~5 per hour per user

packages/ai/src/index.ts
```diff
@@ -1,3 +1,3 @@
-// AI exports
 export * from "./openrouter-client";
 export * from "./rag-service";
+export * from "./crawler";
```
akhileshrangani4: This barrel re-export leaks all the crawler internals (`normalizeUrl`, `isSameDomain`, `fetchRobots`, `USER_AGENT`, etc.) into the top-level `@teachanything/ai` public API.

Since `package.json` already defines a dedicated `"./crawler"` export path, consumers should import from `@teachanything/ai/crawler` directly (which `crawl-processor.ts` already does). This line can be removed.

apps/web/src/app/api/jobs/crawl-discover/route.ts
```diff
@@ -0,0 +1,56 @@
...
+        new Error("Invalid QStash signature"),
+        "Crawl discover job rejected",
+      );
+      return NextResponse.json(
+        { error: "Invalid signature" },
+        { status: 401 },
+      );
+    }
+
+    const { crawlSourceId } = JSON.parse(body);
```
akhileshrangani4: The body is parsed with `JSON.parse` and destructured directly without any validation. `crawlSourceId` goes straight into a DB query without checking it's a valid UUID.

QStash signature verification handles authentication, but defense in depth says validate the payload too:
```typescript
const parsed = z.object({ crawlSourceId: z.string().uuid() }).safeParse(JSON.parse(body));
if (!parsed.success) {
  return NextResponse.json({ error: "Invalid payload" }, { status: 400 });
}
```

Same applies to `crawl-process-page/route.ts`.

apps/web/src/server/routers/crawler/procedures/add-manual-url.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    const [source] = await ctx.db
+      .insert(crawlSources)
+      .values({
+        chatbotId: input.chatbotId,
+        rootUrl: input.url,
+        crawlDepth: 0,
+        maxPages: 1,
+        includePatterns: [],
```
akhileshrangani4, line 84: Unlike `addCrawlSource` which checks for duplicate root URLs, `addManualUrl` creates a new `crawl_sources` row every time without checking if the same URL already exists for this chatbot. Adding the same URL twice creates duplicate sources, pages, and embeddings.

Should check for existing source with the same URL + chatbotId before inserting, same as `addCrawlSource` does.

akhileshrangani4 (changes requested):
Hey! Nice feature overall. The crawler architecture is solid and the split between discovery/processing via QStash is a good call. The UI with live progress polling is a nice touch too.

Left some inline comments below, mostly around security hardening, missing DB indexes, and performance stuff. Copilot already flagged a bunch of the same things (SSRF redirect following, N+1 counts, dev mode finalization, file association scoping) so I tried not to repeat those. Definitely take those Copilot suggestions seriously, they're spot on.

**Build is failing** on Prettier formatting across 13 files. Quick fix: `npm run format` and push.

**Big picture stuff:**

- **SSRF protection needs more work** - Copilot flagged redirect following and DNS resolution already. On top of that, `isUrlSafe()` is only called on the initial root URL, not during BFS crawling. Every URL fetched by `discoverPages` and `fetchAndExtractPage` should also be validated.
- **No rate limiting** on any crawler endpoint. A user can spam crawl sources and burn through embedding API budget fast. Should use the existing Upstash rate limiter pattern from `lib/rate-limit.ts`.
- **Missing DB indexes** on the new tables (see inline comment on the migration).
- **Transactions needed** in `removeCrawlSource` and `processCrawlPage`. Both do multi-step DB mutations without wrapping in a transaction. Partial failures will leave orphaned or stuck data.

See inline comments for specifics!

packages/ai/src/crawler.ts
```diff
@@ -0,0 +1,519 @@
...
+    wordCount: extracted.content.split(/\s+/).length,
+  };
+}
+
+export function isRobotsAllowed(
+  robots: ReturnType<typeof robotsParser>,
+  url: string,
+): boolean {
+  const result = robots.isAllowed(url, USER_AGENT);
+  return result !== false;
```
akhileshrangani4, line 558: `isUrlSafeWithDns()` is defined and exported here but it's never actually called anywhere. Every call site still uses the synchronous `isUrlSafe()`:

- `add-crawl-source.ts` uses `isUrlSafe(input.rootUrl)`
- `add-manual-url.ts` uses `isUrlSafe(input.url)`
- `crawl-processor.ts` uses `isUrlSafe(page.url)`
- BFS discovery loop uses `isUrlSafe(entry.url)`

So the DNS rebinding protection you built doesn't actually run. The procedures and `processCrawlPage` should call `isUrlSafeWithDns()` (the async one) instead of `isUrlSafe()` before any fetch happens.

packages/ai/src/crawler.ts
```diff
@@ -0,0 +1,519 @@
...
+      .replace(/\s+/g, " ")
+      .trim();
+    if (bodyText.length > 50) return { title, content: bodyText };
+    return null;
+  }
+
+  return { title, content: sections.join("\n").trim() };
+}
+
+async function fetchPage(url: string): Promise<{
```
akhileshrangani4: `fetchPage()` still uses `redirect: "follow"` which means redirects are followed blindly. An attacker's URL passes the safety check, then responds with a 302 to `http://169.254.169.254/latest/meta-data/` and the fetch just follows it.

Same issue in the HEAD requests in `add-crawl-source.ts` and `add-manual-url.ts`.

Fix: use `redirect: "manual"`, read the `Location` header, validate each hop with `isUrlSafeWithDns()`, then follow manually. Cap at 5 redirects.

packages/ai/src/crawler.ts
```diff
@@ -0,0 +1,519 @@
...
+  ) {
+    return true;
+  }
+
+  return false;
+}
+
+function isUrlSafe(url: string): boolean {
+  try {
+    const parsed = new URL(url);
```
akhileshrangani4, line 121: `isPrivateIp()` is missing IPv4-mapped IPv6 addresses. DNS can return AAAA records like `::ffff:127.0.0.1` or `::ffff:169.254.169.254`. These bypass the current checks.

Add something like:
```typescript
if (ip.startsWith('::ffff:')) {
  return isPrivateIp(ip.slice(7));
}
```

apps/web/src/lib/crawl-processor.ts
```diff
@@ -0,0 +1,450 @@
...
+        .where(eq(crawledPages.id, crawledPageId));
+      return;
+    }
+
+    await db
+      .update(crawledPages)
+      .set({ status: "processing", updatedAt: new Date() })
+      .where(eq(crawledPages.id, crawledPageId));
+
+    const { fetchAndExtractPage } = await import("@teachanything/ai/crawler");
```
akhileshrangani4: This section still doesn't have transaction wrapping. If embedding fails after the `userFiles` insert + `chatbotFileAssociations` insert, you end up with:
- A `userFiles` row stuck at "processing" forever
- A dangling `chatbotFileAssociation` showing a broken file in the chatbot's list

Also on re-crawl, existing chunks are deleted BEFORE the new embedding attempt (line ~281). If embedding then fails, all the old data for that page is gone and the new data never arrived. That's data loss.

At minimum the catch block should mark the `userFiles` row as "failed" too when `userFileId` exists. Ideally wrap the file creation + association + chunks in a transaction.

apps/web/src/server/routers/crawler/procedures/recrawl.ts
```diff
@@ -0,0 +1,109 @@
...
+      ctx.session.user.id,
+      { action: "recrawl" },
+    );
+    if (!success) {
+      throw new TRPCError({
+        code: "TOO_MANY_REQUESTS",
+        message: "Too many recrawl requests. Please try again later.",
+      });
+    }
+
```
akhileshrangani4, line 26: The active crawl guard is good, but there's no escape hatch if a crawl gets permanently stuck. If all QStash jobs fail silently, the source stays in "crawling" forever. The user can't recrawl (blocked here) and the UI disables the delete button for active crawls. They're trapped.

Consider either:
- Allowing delete on active crawls (just remove the UI disable)
- Adding a timeout: if crawling for more than 2 hours, auto-transition to "failed"
- Or letting recrawl force-reset if the source has been active beyond some threshold

apps/web/src/lib/crawl-processor.ts
```diff
@@ -0,0 +1,450 @@
...
+
+    if (!source) return;
+
+    const [chatbot] = await db
+      .select()
+      .from(chatbots)
+      .where(eq(chatbots.id, source.chatbotId))
+      .limit(1);
+
+    if (!chatbot) return;
```
akhileshrangani4: `fetchRobots(source.rootUrl)` is called inside `processCrawlPage`, which runs as a separate QStash job per page. For a 500-page crawl, that's 500 separate HTTP requests to the same `robots.txt` file.

Consider caching the robots.txt content in `crawl_sources.metadata` during the discovery phase, then reading it from the DB here instead of fetching over HTTP every time.

apps/web/src/server/routers/crawler/procedures/export-json.ts
```diff
@@ -0,0 +1,107 @@
...
+        chunkContent: fileChunks.content,
+        chunkIndex: fileChunks.chunkIndex,
+      })
+      .from(crawledPages)
+      .leftJoin(fileChunks, eq(fileChunks.fileId, crawledPages.userFileId))
+      .where(
+        and(
+          eq(crawledPages.crawlSourceId, input.crawlSourceId),
+          eq(crawledPages.status, "completed"),
+        ),
```
akhileshrangani4, line 29: The `.limit(MAX_EXPORT_PAGES * 50)` operates on the joined rows, not on pages. If a page has 100 chunks it eats 100 rows of that budget. The last page in the result can get its chunks silently truncated, producing incomplete content in the export.

Safer approach: query pages first (limited to 200), then query chunks for those specific page IDs.

apps/web/src/server/routers/crawler/procedures/get-crawl-sources.ts
```diff
@@ -0,0 +1,77 @@
...
+    >();
+
+    for (const row of statusCounts) {
+      if (!countMap.has(row.crawlSourceId)) {
+        countMap.set(row.crawlSourceId, {
+          pending: 0,
+          processing: 0,
+          completed: 0,
+          failed: 0,
+        });
```
akhileshrangani4, line 68: Small gap: the countMap only tracks pending/processing/completed/failed, but the enum also has "blocked" and "skipped". Pages with those statuses get silently dropped from the counts. During an active crawl with blocked pages, the progress bar total (pending + processing + completed + failed) will be less than the actual page count, so it'll never look like it reaches 100%.

akhileshrangani4 (changes requested):
Hey, nice work on the fix commit! You knocked out most of the original feedback. Rate limiting, indexes, transaction on delete, Zod validation, recrawl guard, grouped count query, batched QStash publishing, duplicate URL check, robots.txt enforcement, Radix component fixes, all good.

Still a few things to address though. The big one is the SSRF protection is almost there but not quite wired up. See inline comments.

akhileshrangani4:
Two more things:

**1. Prettier still failing on 3 files.** Run `npx prettier --write .` from the repo root and push. That should clear the CI lint check.

**2. Vercel build is failing with a TypeScript error:**

```
../../packages/ai/src/crawler.ts:2:23
Type error: Could not find a declaration file for module 'jsdom'.
```

Looks like `@types/jsdom` is in `devDependencies` of `packages/ai/package.json`, but Vercel's build environment doesn't install dev deps for workspace packages by default. Two options:

- Move `@types/jsdom` from `devDependencies` to `dependencies` in `packages/ai/package.json`
- Or add a `jsdom.d.ts` declaration file in `packages/ai/src/` with `declare module 'jsdom';`

The first option is the quickest fix.

akhileshrangani4 (approved):

PR #305, opened 2026-06-04
feat(analytics): enhance chatbot analytics with session-level insights

gum1x:
Restored from closed PR #301 (head repo was accidentally deleted).

This is a clean re-creation from a fresh fork of teachanything only. No relation to any agents/research workspace.

Original: https://github.com/akhileshrangani4/teachanything/pull/301


akhileshrangani4:
Post-merge follow-ups from a review pass on this work:

- **#317** — fixes + UI polish: `getCommonQuestions` now returns a correct distinct-question total (the windowed count broke pagination past the last page), `ragUsagePercentage` only counts events that recorded `ragUsed`, Common Questions card is paginated (5/page), and Session Lengths spans full width to drop the empty grid column.
- **#315** — tracks semantic grouping for common questions (current grouping is exact-match on normalized text).

Copilot's other notes left as-is on purpose: the timezone day/week keys are a no-op under Supabase UTC, plain `CREATE INDEX` is fine at this table size (Drizzle migrations run in a transaction, so `CONCURRENTLY` isn't available), and the stored `question` excerpt is kept since the full text already lives in `messages.content`.

PR #306, opened 2026-06-04
feat(files): Add OCR support for image uploads and scanned PDFs

gum1x:
Restored from closed PR #293 (head repo was accidentally deleted).

This is a clean re-creation from a fresh fork of teachanything only.

Original: https://github.com/akhileshrangani4/teachanything/pull/293

### DIFF
--- a/apps/web/src/components/dashboard/files/FileStatusBadge.tsx
+++ b/apps/web/src/components/dashboard/files/FileStatusBadge.tsx
@@ -20,6 +20,8 @@ interface ProcessingProgress {
   percentage: number;
   currentChunk?: number;
   totalChunks?: number;
+  currentPage?: number;
+  totalPages?: number;
   startedAt?: string;
   lastUpdatedAt?: string;
 }
@@ -173,6 +175,11 @@ export function FileStatusBadge({
               Chunk {progress.currentChunk || 0} of {progress.totalChunks}
             </p>
           )}
+          {progress.totalPages && (
+            <p className="text-xs">
+              Page {progress.currentPage || 0} of {progress.totalPages}
+            </p>
+          )}
           <p className="text-xs">Progress: {progress.percentage.toFixed(0)}%</p>
         </div>
       );
@@ -235,6 +242,13 @@ export function FileStatusBadge({
                   progress.totalChunks > 0 && (
                     <span className="text-[10px] text-muted-foreground tabular-nums whitespace-nowrap">
                       {progress.currentChunk || 0}/{progress.totalChunks}
+                    </span>
+                  )}
+                {progress.stage === "extracting" &&
+                  progress.totalPages &&
+                  progress.totalPages > 0 && (
+                    <span className="text-[10px] text-muted-foreground tabular-nums whitespace-nowrap">
+                      Page {progress.currentPage || 0}/{progress.totalPages}
                     </span>
                   )}
                 <span className="text-xs font-medium text-foreground tabular-nums min-w-[35px] text-right">
--- a/apps/web/src/components/dashboard/files/FileStatusBadge.tsx
+++ b/apps/web/src/components/dashboard/files/FileStatusBadge.tsx
@@ -175,7 +175,7 @@ export function FileStatusBadge({
               Chunk {progress.currentChunk || 0} of {progress.totalChunks}
             </p>
           )}
-          {progress.totalPages && (
+          {progress.stage === "extracting" && progress.totalPages && (
             <p className="text-xs">
               Page {progress.currentPage || 0} of {progress.totalPages}
             </p>
--- a/apps/web/src/components/dashboard/files/file-constants.ts
+++ b/apps/web/src/components/dashboard/files/file-constants.ts
@@ -1,3 +1,8 @@
+import {
+  ALLOWED_EXTENSIONS,
+  FILE_TYPE_DISPLAY_NAMES,
+} from "@/lib/upload-file-types";
+
 // Get max file size from environment variable (client-accessible via NEXT_PUBLIC_ prefix)
 // Falls back to 50MB if not set
 const getMaxFileSizeMB = (): number => {
@@ -7,42 +12,16 @@ const getMaxFileSizeMB = (): number => {
 
 export const MAX_FILE_SIZE = getMaxFileSizeMB() * 1024 * 1024;
 export const MAX_FILE_NAME_LENGTH = 255;
-export const ALLOWED_FILE_TYPES = [
-  "application/pdf",
-  "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
-  "application/msword",
-  "application/vnd.openxmlformats-officedocument.presentationml.presentation",
-  "text/plain",
-  "text/markdown",
-  "application/json",
-  "text/csv",
-] as const;
-
-export const ALLOWED_EXTENSIONS = [
-  ".pdf",
-  ".doc",
-  ".docx",
-  ".pptx",
-  ".txt",
-  ".md",
-  ".markdown",
-  ".json",
-  ".csv",
-] as const;
-
-// User-friendly file type names for error messages
-export const FILE_TYPE_DISPLAY_NAMES: Record<string, string> = {
-  "application/pdf": "PDF",
-  "application/vnd.openxmlformats-officedocument.wordprocessingml.document":
-    "Word (.docx)",
-  "application/msword": "Word (.doc)",
-  "application/vnd.openxmlformats-officedocument.presentationml.presentation":
-    "PPTX",
-  "text/plain": "Text",
-  "text/markdown": "Markdown",
-  "application/json": "JSON",
-  "text/csv": "CSV",
-};
+export {
+  ALLOWED_FILE_TYPES,
+  ALLOWED_EXTENSIONS,
+  FILE_INPUT_ACCEPT,
+  FILE_TYPE_DISPLAY_NAMES,
+  OCR_MAX_IMAGE_SIZE_BYTES,
+  OCR_MAX_IMAGE_SIZE_MB,
+  inferSupportedFileType,
+  isOCRImageFileType,
+} from "@/lib/upload-file-types";
 
 // Helper to get user-friendly file type name
 export const getFileTypeDisplayName = (mimeType: string): string => {
--- a/apps/web/src/lib/file-processor.ts
+++ b/apps/web/src/lib/file-processor.ts
@@ -8,7 +8,19 @@ import { EMBEDDING_MODEL } from "@teachanything/ai/models";
 import { env } from "./env";
 import { logInfo, logError } from "./logger";
 
-const EXTRACTION_TIMEOUT_MS = 60_000;
+const EXTRACTION_TIMEOUT_MS = 30 * 60_000;
+
+/**
+ * Processing batch sizes - tuned for performance and rate limiting
+ */
+const PROCESSING_CONFIG = {
+  /** Chunks per embedding batch (50 to avoid OpenRouter rate limits) */
+  EMBEDDING_BATCH_SIZE: 50,
+  /** Chunk records per database insert batch (500 for efficient bulk insert) */
+  DB_INSERT_BATCH_SIZE: 500,
+} as const;
+
+type FileMetadata = NonNullable<typeof userFiles.$inferSelect.metadata>;
 
 /**
  * Sanitize error messages before storing in metadata visible to users.
@@ -16,27 +28,27 @@ const EXTRACTION_TIMEOUT_MS = 60_000;
  */
 function sanitizeProcessingError(error: unknown): string {
   const msg = error instanceof Error ? error.message : String(error);
+  const errorName = error instanceof Error ? error.name : "";
   if (msg.includes("timed out")) return "File processing timed out";
   if (msg.includes("Unsupported file type")) return msg;
   if (msg.includes("no readable text")) return msg;
+  if (msg.includes("Invalid image format")) return "Invalid image format";
+  if (msg.includes("Image exceeds OCR size limit"))
+    return "Image exceeds OCR size limit";
+  if (msg.includes("Image dimensions exceed"))
+    return "Image dimensions exceed maximum limit";
+  if (msg.includes("too many pages for OCR")) return msg;
+  if (msg.includes("too large to render for OCR")) return msg;
   if (msg.includes("Invalid PDF")) return "Invalid PDF format";
   if (msg.includes("embedding") && msg.includes("dimension"))
     return "Embedding dimension mismatch";
+  if (
+    msg.includes("Extraction aborted") ||
+    msg.includes("timed out") ||
+    errorName === "AbortError"
+  )
+    return "File processing timed out";
   return "File processing failed due to an internal error";
-}
-
-function withTimeout<T>(
-  promise: Promise<T>,
-  ms: number,
-  message: string,
-): Promise<T> {
-  let timeoutId: ReturnType<typeof setTimeout>;
-  const timeout = new Promise<never>((_, reject) => {
-    timeoutId = setTimeout(() => reject(new Error(message)), ms);
-  });
-  return Promise.race([promise, timeout]).finally(() =>
-    clearTimeout(timeoutId),
-  );
 }
 
 /**
@@ -48,46 +60,64 @@ async function updateProgress(
   percentage: number,
   currentChunk?: number,
   totalChunks?: number,
+  currentPage?: number,
+  totalPages?: number,
 ) {
-  const now = new Date().toISOString();
+  try {
+    const now = new Date().toISOString();
 
-  // Get current file to preserve existing metadata
-  const [currentFile] = await db
-    .select()
-    .from(userFiles)
-    .where(eq(userFiles.id, fileId))
-    .limit(1);
+    const [currentFile] = await db
+      .select()
+      .from(userFiles)
+      .where(eq(userFiles.id, fileId))
+      .limit(1);
 
-  const existingMetadata = currentFile?.metadata || {};
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const { error: _prevError, ...cleanMetadata } = existingMetadata;
-  const startedAt = existingMetadata?.processingProgress?.startedAt || now;
+    const existingMetadata: FileMetadata = currentFile?.metadata || {};
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { error: _prevError, ...cleanMetadata } = existingMetadata;
+    const startedAt = existingMetadata?.processingProgress?.startedAt || now;
 
-  await db
-    .update(userFiles)
-    .set({
-      processingStatus: "processing",
-      metadata: {
-        ...cleanMetadata,
-        processingProgress: {
-          stage,
-          percentage: Math.min(100, Math.max(0, percentage)),
-          currentChunk,
-          totalChunks,
-          startedAt,
-          lastUpdatedAt: now,
+    await db
+      .update(userFiles)
+      .set({
+        processingStatus: "processing",
+        metadata: {
+          ...cleanMetadata,
+          processingProgress: {
+            stage,
+            percentage: Math.min(100, Math.max(0, percentage)),
+            currentChunk,
+            totalChunks,
+            currentPage,
+            totalPages,
+            startedAt,
+            lastUpdatedAt: now,
+          },
         },
-      },
-    })
-    .where(eq(userFiles.id, fileId));
+      })
+      .where(
+        and(
+          eq(userFiles.id, fileId),
+          eq(userFiles.processingStatus, "processing"),
+        ),
+      );
 
-  logInfo(`File processing progress: ${stage} ${percentage}%`, {
-    fileId,
-    stage,
-    percentage,
-    currentChunk,
-    totalChunks,
-  });
+    logInfo(`File processing progress: ${stage} ${percentage}%`, {
+      fileId,
+      stage,
+      percentage,
+      currentChunk,
+      totalChunks,
+      currentPage,
+      totalPages,
+    });
+  } catch (err) {
+    logError(err, "Failed to update processing progress (non-fatal)", {
+      fileId,
+      stage,
+      percentage,
+    });
+  }
 }
 
 /**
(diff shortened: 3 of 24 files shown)


apps/web/src/lib/file-processor.ts
```diff
@@ -8,37 +8,49 @@ import { EMBEDDING_MODEL } from "@teachanything/ai/models";
 import { env } from "./env";
 import { logInfo, logError } from "./logger";
 
-const EXTRACTION_TIMEOUT_MS = 60_000;
+const EXTRACTION_TIMEOUT_MS = 30 * 60_000;
```
akhileshrangani4: **P0 + P1 (timeout cluster).** This is now 30 min, but the QStash route `apps/web/src/app/api/jobs/process-file/route.ts` has no `maxDuration` export (the crawl routes both use `maxDuration = 300`). So:

1. The function is killed at the platform default well before this timer fires — OCR jobs that need real time never finish.
2. A killed/timed-out job leaves the row stuck in `processing`; the atomic status guard turns every QStash retry into a silent no-op and there's no stale-job reaper.
3. The AbortController abort doesn't actually interrupt an in-flight Tesseract `recognize()` / pdfjs `render()` — tesseract.js doesn't honor AbortSignal, so work runs on until the inner 120s `RECOGNITION_TIMEOUT_MS` or `terminate()`.

Fix: add `export const maxDuration = <plan ceiling>` to the route, set `EXTRACTION_TIMEOUT_MS` ~30-60s under it (minus the download/chunk/embed/store budget), make the inner per-call timeout the real cancellation bound, and let the status guard reclaim `processing` rows whose `lastUpdatedAt` is older than timeout+buffer.

apps/web/src/lib/file-processor.ts
```diff
@@ -48,46 +60,64 @@ async function updateProgress(
...
-          stage,
-          percentage: Math.min(100, Math.max(0, percentage)),
-          currentChunk,
-          totalChunks,
-          startedAt,
-          lastUpdatedAt: now,
+  try {
+    const now = new Date().toISOString();
+
+    const [currentFile] = await db
```
akhileshrangani4: **P1 (perf).** `updateProgress` re-SELECTs the row just to read `startedAt` before every UPDATE — two round-trips per tick. A 30-page scanned PDF fires ~69 of these = ~138 round-trips. Capture `startedAt` once at job start and pass it in; drop the SELECT entirely (the conditional UPDATE already guards on `processingStatus = 'processing'`).

apps/web/src/lib/file-processor.ts
```diff
@@ -203,12 +256,43 @@ export async function processFile(params: {
 
     // Stage 2: Extract text content (10-30%)
     await updateProgress(fileId, "extracting", 10);
-    const ragService = createRAGService();
-    const content = await withTimeout(
-      ragService.extractContent(buffer, file.fileType),
-      EXTRACTION_TIMEOUT_MS,
-      `File extraction timed out after ${EXTRACTION_TIMEOUT_MS / 1000}s`,
-    );
+    let isExtractionActive = true;
```
akhileshrangani4: **P3 (cleanup).** `isExtractionActive` guards a path that can't happen: the `onProgress` callback only runs inside `extractContent`, which has already settled by the time the `finally` sets the flag false. The `if (!isExtractionActive) return` can never fire — safe to remove the flag and the guard.

packages/ai/src/rag-service.ts
```diff
@@ -117,37 +225,394 @@ export class RAGService {
...
+    buffer: Buffer,
+    mimeType?: string,
+    signal?: AbortSignal,
+  ): Promise<string> {
+    if (signal?.aborted) throw signal.reason;
+    this.validateImageBuffer(buffer, mimeType);
+
+    const { loadImage } = await import("@napi-rs/canvas");
+    try {
+      const img = await loadImage(buffer);
```
akhileshrangani4: **P1 (security — decompression bomb).** `validateImageBuffer` caps *compressed* bytes at 25MB and checks magic bytes, but `loadImage(buffer)` here does a full in-memory decode **before** the `width*height > MAX_RENDERED_PAGE_PIXELS` check below. A 25MB highly-compressed PNG/TIFF decodes to a huge bitmap and can OOM the worker. Note the PDF-render path (`extractTextFromRenderedPDFPage`) correctly bounds pixels *before* rasterizing — the direct-image path should do the same: probe header dimensions before decode, or lower the byte cap to bound worst-case decoded memory.

packages/ai/src/rag-service.ts
```diff
@@ -3,12 +3,82 @@ import { logWarn, logError } from "@teachanything/logger";
 import type { OpenRouterClient } from "./openrouter-client";
 import { CHARS_PER_TOKEN } from "./token-budget";
 
+/**
+ * OCR and extraction configuration
+ */
+const OCR_CONFIG = {
+  /** Max image size before rejecting (25MB) */
+  MAX_IMAGE_BYTES: 25 * 1024 * 1024,
```
akhileshrangani4: **P2 (DRY / drift risk).** This 25MB limit is duplicated — `OCR_MAX_IMAGE_SIZE_BYTES` in `apps/web/src/lib/upload-file-types.ts` is the same value used by upload validation. Two sources of truth for the same security limit will drift. Import the shared constant here instead of redefining it.

packages/ai/src/rag-service.ts
```diff
@@ -3,12 +3,82 @@ import { logWarn, logError } from "@teachanything/logger";
...
+  /** Multi-page PDFs below this extracted text length are likely scanned. */
+  SCANNED_PDF_TEXT_THRESHOLD: 50,
+  /** Individual fallback-rendered pages below this length are OCR candidates. */
+  PDF_PAGE_OCR_TEXT_THRESHOLD: 10,
+  /** Canvas render scale for PDF→image OCR (2x for quality) */
+  PDF_RENDER_SCALE: 2,
+  /** Avoid unbounded OCR work for huge scanned documents */
+  MAX_PDF_OCR_PAGES: 30,
```
akhileshrangani4: **P2 (reliability).** 30 pages x the 120s per-page `RECOGNITION_TIMEOUT_MS` = up to ~60 min of OCR, which is double the 30-min `EXTRACTION_TIMEOUT_MS`. The page-count cap and the time budget aren't mutually consistent. Add a cumulative OCR wall-clock budget that aborts once total OCR time is exceeded, or lower `MAX_PDF_OCR_PAGES` / `MAX_RENDERED_PAGE_PIXELS`.

packages/ai/src/rag-service.ts
```diff
@@ -117,37 +225,394 @@ export class RAGService {
...
+      parsed.pageCount > 1 &&
+      parsed.text.length < OCR_CONFIG.SCANNED_PDF_TEXT_THRESHOLD
+    );
+  }
 
-      if (!sanitizedText) {
-        throw new Error("PDF contains no readable text content");
+  private async loadPDFDocument(buffer: Buffer): Promise<PdfDocument> {
+    const pdfjsLib =
+      (await import("pdfjs-dist/legacy/build/pdf.mjs")) as unknown as {
```
akhileshrangani4: **P2 (type safety).** `as unknown as { getDocument: ... }` throws away all type checking on the pdfjs API — a breaking change in pdfjs won't surface as a compile error. Import the type and cast narrowly instead: `import type { getDocument as GetDocumentFn } from 'pdfjs-dist'` then `as { getDocument: typeof GetDocumentFn }`. (AGENTS.md: avoid `any`/`unknown` leaks.)

packages/ai/src/rag-service.ts
```diff
@@ -117,37 +225,394 @@ export class RAGService {
...
       const pdfHeader = buffer.subarray(0, 4).toString();
       if (pdfHeader !== "%PDF") {
         throw new Error(
           `Invalid PDF format: expected PDF header, got "${pdfHeader}"`,
         );
       }
 
-      // Dynamic import to avoid build-time execution
-      const pdfParse = (await import("pdf-parse")).default;
+      const parsed = await this.extractPDFText(buffer);
```
akhileshrangani4: **P2 (perf).** Scanned PDFs get parsed twice — `pdf-parse` here, then `pdfjs-dist` re-parses the same buffer in the OCR fallback. pdfjs already gives you `numPages` and per-page text, which is all `shouldUsePDFOCRFallback` needs. Consider using pdfjs as the single parser and dropping pdf-parse for this path (also kills the `new Uint8Array(buffer)` copy).

packages/ai/src/rag-service.ts
```diff
@@ -117,37 +225,394 @@ export class RAGService {
...
+      );
+      return "";
+    }
+
+    try {
+      return await this.extractTextFromTrustedImage(buffer, signal);
+    } catch (error) {
+      if (
+        error instanceof Error &&
+        error.message.includes("Image contains no readable text content")
```
akhileshrangani4: **P2 (reliability — silent data loss).** Failed pages `logWarn` and return `""`, and `pageTexts.filter(Boolean)` quietly drops them. If *every* page fails to OCR, you fall through to the generic "no readable text content" error with no signal that OCR actually crashed — and a partially-failed PDF silently produces a chunk set missing that content. Track a failed-page counter; if it equals `document.numPages`, throw something descriptive instead.

packages/ai/src/rag-service.ts
```diff
@@ -117,37 +225,394 @@ export class RAGService {
...
+        canvasContext: context,
+        canvas,
+        viewport,
+      }).promise;
+
+      buffer = canvas.toBuffer("image/png");
+    } catch (error) {
+      if (
+        error instanceof Error &&
+        error.message.includes("too large to render for OCR")
```
akhileshrangani4: **P2 (maintainability).** Control flow keyed on `error.message.includes('too large to render for OCR')` (and similar in `extractTextFromImage` / `sanitizeProcessingError`) is fragile — rephrasing a message silently changes behavior. Introduce small named error subclasses (e.g. `OcrSizeError`, `OcrNoTextError`) with a `code` and switch on `instanceof`.

packages/ai/src/rag-service.ts
```diff
@@ -397,7 +864,13 @@ export class RAGService {
       this.encoder.free();
     }
     this.encoder = null;
-    this.encoderInitialized = false;
+    this.encoderInitPromise = null;
+
+    const workerPromise = this.ocrWorkerInitPromise;
+    this.ocrWorkerInitPromise = null;
+    if (workerPromise) {
+      void workerPromise.then((w) => w.terminate()).catch(() => {});
```
akhileshrangani4: **P2 (observability).** `.catch(() => {})` swallows worker-termination failures with zero trace. At least `logWarn` it — a wedged native worker that fails to terminate is exactly the thing you'll want a breadcrumb for. (AGENTS.md: no silent failures.)

apps/web/src/server/routers/files/validation.ts
```diff
@@ -125,10 +117,18 @@ export function validateExtensionMatchesMimeType(
...
-  const extension = fileNameLower.substring(fileNameLower.lastIndexOf(".") + 1);
+  const lastDot = fileNameLower.lastIndexOf(".");
+  const extension = lastDot >= 0 ? fileNameLower.substring(lastDot + 1) : "";
   const validMimeTypes = EXTENSION_MIME_MAP[extension];
 
-  if (extension && validMimeTypes && !validMimeTypes.includes(fileType)) {
+  if (!extension || !validMimeTypes) {
+    throw new TRPCError({
+      code: "BAD_REQUEST",
+      message: `File extension${extension ? ` (.${extension})` : ""} is not supported.`,
```
akhileshrangani4, line 127: **P2 (contract change).** This is a behavior change: `validateExtensionMatchesMimeType` used to silently pass for unknown/missing extensions, now it throws BAD_REQUEST. The normal UI path always supplies an inferred extension so no user regression, but a direct tRPC caller with an extensionless filename that used to work now gets rejected (at both create-upload-url and finalize-upload). Intentional hardening is fine — just add a changelog/docs note so it's not a surprise. The tests were updated to match without flagging it as a contract change.

apps/web/src/server/routers/files/validation.ts
```diff
@@ -1,54 +1,33 @@
...
-  "application/pdf",
-  "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
-  "application/msword",
-  "application/vnd.openxmlformats-officedocument.presentationml.presentation",
-  "text/plain",
-  "text/markdown",
-  "application/json",
-  "text/csv",
-] as const;
+export const SUPPORTED_FILE_TYPES = ALLOWED_FILE_TYPES;
```
akhileshrangani4, line 15: **P2 (cleanup).** `SUPPORTED_FILE_TYPES` is just an alias of `ALLOWED_FILE_TYPES` now — two exported names for the same array. Drop the alias and import `ALLOWED_FILE_TYPES` directly in callers/tests. Zero behavior change.

apps/web/src/lib/upload-file-types.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export function getFileExtension(fileName: string): string | null {
+  const lastDot = fileName.lastIndexOf(".");
+  if (lastDot <= 0 || lastDot === fileName.length - 1) {
+    return null;
+  }
+  return fileName.slice(lastDot).toLowerCase();
+}
+
+export function inferSupportedFileType(
```
akhileshrangani4, line 95: **P2 (agent-native / API parity).** `inferSupportedFileType` only runs client-side (here it's called from `direct-upload.ts` / the dialog). A raw tRPC caller — or an agent hitting `files.createUploadUrl` directly — has to know `.tiff` -> `image/tiff` itself or it gets BAD_REQUEST. This fn is pure and server-safe; consider calling it inside `createUploadUrlProcedure` before `validateFileType` so every caller gets the same inference the UI does.

apps/web/src/lib/upload-file-types.ts
```diff
@@ -0,0 +1,120 @@
...
+  "image/png": "PNG image",
+  "image/webp": "WEBP image",
+  "image/tiff": "TIFF image",
+  "text/plain": "Text",
+  "text/markdown": "Markdown",
+  "application/json": "JSON",
+  "text/csv": "CSV",
+};
+
+export function getFileExtension(fileName: string): string | null {
```
akhileshrangani4, line 87: **P2 (tests).** `getFileExtension` and `isOCRImageFileType` are new pure functions feeding security-relevant type inference, but neither has direct tests. Worth covering the edge cases — `getFileExtension` for `.gitignore` (leading dot), `noext`, trailing dot; `isOCRImageFileType` false for `application/pdf`.

apps/web/src/components/dashboard/files/file-constants.ts
```diff
@@ -7,42 +12,16 @@ const getMaxFileSizeMB = (): number => {
...
-    "Word (.docx)",
-  "application/msword": "Word (.doc)",
-  "application/vnd.openxmlformats-officedocument.presentationml.presentation":
-    "PPTX",
-  "text/plain": "Text",
-  "text/markdown": "Markdown",
-  "application/json": "JSON",
-  "text/csv": "CSV",
-};
+export {
```
akhileshrangani4, line 15: **P2 (maintainability).** This file is now basically a pure re-export shim pointing at `@/lib/upload-file-types` — dead indirection. Either point the dashboard component imports straight at the lib module and delete this, or keep only the component-layer helpers that have no home in the lib.

packages/ai/src/openrouter-client.ts
```diff
@@ -78,7 +89,7 @@ export class OpenRouterClient {
     messages: Array<{ role: "system" | "user" | "assistant"; content: string }>;
     temperature?: number;
     maxTokens?: number;
-  }) {
+  }): Promise<OpenRouterStreamTextResult> {
```
akhileshrangani4, line 89: Note (not blocking). Narrowing the return type to just `{ fullStream }` is fine for today — the only caller (`chat.ts:264`) consumes `result.fullStream` and nothing else. Heads up that the runtime object still has `.text`, `.usage`, `.textStream`, etc., so a future caller reaching for those will get a confusing compile error against a value that actually has them. Re-exporting the SDK's `StreamTextResult` type would avoid that trap.

packages/ai/package.json
```diff
@@ -14,12 +14,14 @@
   },
   "dependencies": {
     "@ai-sdk/openai": "^3.0.53",
-    "@teachanything/logger": "*",
     "@langchain/community": "^1.1.27",
     "@langchain/core": "^1.1.42",
     "@langchain/textsplitters": "^1.0.0",
     "@mozilla/readability": "^0.6.0",
+    "@napi-rs/canvas": "^0.1.100",
```
akhileshrangani4, line 22: **P2 (confirm intent).** Three runtime deps added (`@napi-rs/canvas`, `pdfjs-dist`, `tesseract.js`) plus a Node engine bump to `>=20.19.0`. All justified for OCR, and nicely the `serverExternalPackages` list was updated to match (per the repo's own ADR) — just flagging for explicit sign-off, and double-check CI/CD runners are on Node >=20.19.0 so the engine bump doesn't bite.

apps/web/src/components/dashboard/files/FileStatusBadge.tsx
```diff
@@ -173,6 +175,11 @@ export function FileStatusBadge({
               Chunk {progress.currentChunk || 0} of {progress.totalChunks}
             </p>
           )}
+          {progress.totalPages && (
```
akhileshrangani4: Minor nit. The inline row (a bit below) gates the page count on `progress.stage === 'extracting'`, but this tooltip block only checks `progress.totalPages`. Once totalPages is set it'll keep showing the page line into later stages here. Gate it the same way as the inline row for consistency.

packages/ai/src/__tests__/ocr-extraction.test.ts
```diff
@@ -0,0 +1,484 @@
...
+  });
+
+  it("reuses the same worker across multiple calls", async () => {
+    recognizeMock.mockResolvedValue({ data: { text: "text" } });
+    await service.extractTextFromImage(pngBuffer, "image/png");
+    await service.extractTextFromImage(pngBuffer, "image/png");
+    expect(createWorkerMock).toHaveBeenCalledTimes(1);
+  });
+
+  it("rejects images whose header does not match the declared MIME type", async () => {
```
akhileshrangani4, line 232: **P2 (test gaps).** A few holes in otherwise great coverage:

- Magic-byte mismatch is only tested PNG-as-JPEG. `detectImageMimeType` has 4 independent header branches — add webp (RIFF), tiff (II*/MM), jpeg-as-png, and the unknown-header case.
- The "terminates worker on cleanup during init" test (~line 189) passes for the wrong reason: `recognizeMock` isn't set so `result.data.text` throws a TypeError — the rejection isn't proving interruption. Set a valid recognize result and assert `recognizeMock` was NOT called.
- The blank-image path (`extractTextFromImage` -> "Image contains no readable text content") isn't tested for the direct-image case (the PDF-page analog is).

packages/ai/src/__tests__/rag-service.test.ts
```diff
@@ -176,6 +232,43 @@ describe("RAGService", () => {
...
+      expect(result).toBe("mocked ocr text");
+    });
+
+    it("extracts text from PDF buffer", async () => {
+      const buffer = Buffer.from("%PDF-1.4 mock data");
+      const result = await service.extractContent(buffer, "application/pdf");
+      expect(result).toBe("mocked pdf text");
+    });
+
+    it("respects abort signals during extraction", async () => {
```
akhileshrangani4, line 262: **P2 (test gap).** Abort is only tested with a pre-aborted signal, so the per-page `signal?.aborted` checks inside `extractPDFWithOCRFallback` are never exercised. Add a test that fires `controller.abort()` after the page loop starts (e.g. mock getPage to abort on page 2 of 5), then assert it rejects with the abort reason AND `destroy()` still ran once (proving the `finally` fired).

akhileshrangani4 (changes requested):
Hey, did a deep pass on this with a multi-agent review (correctness, security, reliability, perf, api-contract, testing, maintainability, standards). Really solid feature overall and the test coverage is genuinely impressive for something this size. But there's one blocker that makes the whole thing a no-op in production, plus a handful of things worth tightening before merge.

**The big one (P0):** the extraction timeout got bumped from 60s to 30 minutes, but `apps/web/src/app/api/jobs/process-file/route.ts` doesn't export a `maxDuration`. Both sibling job routes (`crawl-discover`, `crawl-process-page`) set `maxDuration = 300`. Without it, the function gets killed at the platform default (~300s) long before the 30-min abort timer, the page limits, or the pixel guards can ever fire. So every scanned PDF or large image that actually needs real OCR time gets killed mid-run and the row is left stuck in `processing` forever (the atomic guard makes QStash retries silent no-ops, and there's no stale-job reaper). Left inline comments on the trigger line.

**Counts:** 1 P0, 4 P1, ~18 P2, a few P3. Inline comments below have the details + suggested fixes.

A couple of things that don't fit on a single line:

- **rag-service.ts is ~882 lines now** (AGENTS.md targets <300). The OCR stuff (worker mgmt, image validation, PDF rendering, magic-byte detection) is cohesive enough to pull into its own `ocr-service.ts` that RAGService delegates to. Would bring the file back under control.
- **Worst-case OCR math doesn't close:** `MAX_PDF_OCR_PAGES = 30` x 120s per-page recognition timeout = up to 60 min, which is double the 30-min extraction budget (which itself can't run, see P0). Worth a cumulative OCR wall-clock budget, or lower the page/pixel caps.
- **Pre-existing (not from this PR, just noting):** the server trusts the client-declared `fileType`; magic-byte verification now covers images + PDF but not docx/pptx content. Separate hardening pass someday.

One correction to my own notes: the `streamText` return-type narrowing looked scary but the only caller (`chat.ts:264`) just consumes `result.fullStream`, so nothing actually breaks today. Flagged it inline as a future-caller trap only, not a blocker.

Nice work, just needs the maxDuration fix to actually function and a few rough edges sanded. 🙏

_One more perf note that's pre-existing (not introduced here, so no inline): `countTokens` runs inside a `Promise.all` over every chunk in `file-processor.ts`, firing all the tiktoken WASM encodes concurrently. On big files that saturates the event loop — worth making sequential or folding into the embedding batch loop at some point._

PR #307, opened 2026-06-04
feat(chat): add multilingual voice input via Whisper

gum1x:
Restored from closed PR #290 (head repo was accidentally deleted).

This is a clean re-creation from a fresh fork of teachanything only. The ocr-extraction test removal is included to match original scope.

Original: https://github.com/akhileshrangani4/teachanything/pull/290


apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  const lengthError = checkContentLength(request);
+  if (lengthError) return lengthError;
+
+  const authResult = await resolveAuth(request);
+  if ("error" in authResult) return authResult.error;
+  const { identifier, limiter, logContext, inferredChatbotId, userId } =
+    authResult;
+
+  const { success, reset } = await checkRateLimit(limiter, identifier, {
```
akhileshrangani4: **Blocker: fail-open rate limit on a paid endpoint.** `checkRateLimit` returns `success: true` when the limiter is null (Redis down). On the unauthenticated shareToken surface that turns a Redis outage into an unmetered Whisper proxy, anyone can run up the OpenAI bill. Use a fail-closed path here (`requireRateLimit`, or explicitly deny when `limiter === null`) for the public surface. If fail-open was a deliberate availability call, can you note why? For a paid external API I think fail-closed is the safer default.

apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
...
+  const shareToken = new URL(request.url).searchParams.get("shareToken");
+  if (!shareToken) {
+    return { error: jsonError("Unauthorized", "unauthorized", 401) };
+  }
+
+  // Compute the rate-limit identifier from shareToken + IP before doing
+  // any DB lookups, so rate-limiting happens first and expensive lookups
+  // are protected.
+  const clientIp =
+    request.headers.get("x-forwarded-for")?.split(",")[0]?.trim() ?? "unknown";
```
akhileshrangani4: **Blocker: spoofable rate-limit bucket.** `x-forwarded-for.split(",")[0]` is the leftmost hop, which the client controls. A caller just rotates that header to get unlimited buckets per shareToken, which defeats the public limiter entirely (and pairs badly with the fail-open issue above). Derive the IP from the trusted edge hop (rightmost appended XFF, or `x-real-ip`), and consider a per-shareToken global cap so total spend per token is bounded even under IP rotation.

apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
...
+ * identifier is computed before DB lookup so rate-limit checks happen first.
+ */
+async function resolveAuth(
+  request: NextRequest,
+): Promise<AuthContext | { error: NextResponse<ErrorResponse> }> {
+  const hasCookies = !!request.headers.get("cookie");
+
+  if (hasCookies) {
+    const session = await auth.api.getSession({ headers: request.headers });
+    if (session?.user) {
```
akhileshrangani4, line 85: **Blocker (please confirm intent): approval/ban gate skipped.** This accepts any `session?.user`, but `protectedProcedure` everywhere else also requires `status === "approved"` (admins bypass). So an unapproved or banned logged-in user gets the higher per-user transcription bucket on the paid endpoint. Anonymous users can already transcribe via shareToken, so this is partly a consistency question, but a banned user especially shouldn't. Mirror the `protectedProcedure` check here, or document that voice is intentionally ungated.

apps/web/src/components/chat/messages/VoiceInputButton.tsx
```diff
@@ -0,0 +1,261 @@
...
+            ? "ogg"
+            : "webm";
+        form.append("audio", blob, `recording.${ext}`);
+        if (chatbotId) form.append("chatbotId", chatbotId);
+
+        const url = shareToken
+          ? `/api/transcribe?shareToken=${encodeURIComponent(shareToken)}`
+          : "/api/transcribe";
+
+        const res = await fetch(url, { method: "POST", body: form });
```
akhileshrangani4: **Blocker: fetch has no AbortController and outlives unmount.** A Whisper call can take 20-60s. If the component unmounts (navigation, chatbot switch) while it's in flight, the `finally` still calls `setIsTranscribing(false)` on a dead component, and the transcript is silently dropped. Add an AbortController ref, abort it in a `useEffect` cleanup, guard the `finally` with a mounted check, and add a client-side timeout (~100s, just above the 90s server cap) so a stalled response doesn't pin the spinner forever.

apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  // Compute the rate-limit identifier from shareToken + IP before doing
+  // any DB lookups, so rate-limiting happens first and expensive lookups
+  // are protected.
+  const clientIp =
+    request.headers.get("x-forwarded-for")?.split(",")[0]?.trim() ?? "unknown";
+  const identifier = `${shareToken}:${clientIp}`;
+
+  // Validate the shareToken exists and is enabled
+  const [chatbot] = await db
```
akhileshrangani4: **Blocker: DB errors break the JSON error contract.** This query (inside `resolveAuth`) and the `findOwnedChatbotId` call at line 185 both run before the `try` at line 190. If the DB throws (pool exhaustion, dropped connection), Next returns an HTML 500 instead of the machine-readable `{ error, code }` shape every other path returns. Wrap both in try/catch and return `jsonError(..., "internal_error", 500)`.

apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
+import { NextRequest, NextResponse } from "next/server";
+import { auth } from "@/lib/auth";
+import { db } from "@teachanything/db";
+import { chatbots, analytics } from "@teachanything/db/schema";
+import { eq, and } from "drizzle-orm";
+import { findOwnedChatbotId } from "@/server/routers/chatbot";
```
akhileshrangani4: **Blocker: cross-layer leak.** The route imports `findOwnedChatbotId` from a tRPC router file, and that helper's `ChatbotDb` type is derived from `protectedProcedure.query`'s internals (`chatbot.ts:22`). Any change to the tRPC procedure type silently breaks this route. Move `findOwnedChatbotId` into a dedicated module like `server/queries/chatbot.ts` with a concrete Drizzle `db` type from `@teachanything/db`, and import it from there in both the router and the route.

apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
...
+      };
+    }
+  }
+
+  const shareToken = new URL(request.url).searchParams.get("shareToken");
+  if (!shareToken) {
+    return { error: jsonError("Unauthorized", "unauthorized", 401) };
+  }
+
+  // Compute the rate-limit identifier from shareToken + IP before doing
```
akhileshrangani4: The comment says rate-limiting happens first so the lookup is protected, but the lookup is inside `resolveAuth` (called at line 130), and `checkRateLimit` runs after at line 135. So the DB lookup is not actually protected by the limiter. Either move the rate-limit check before the shareToken lookup, or fix the comment so it doesn't claim a guarantee that isn't there.

apps/web/src/components/chat/messages/ChatInput.tsx
```diff
@@ -58,6 +80,21 @@ export function ChatInput({
     }
   };
 
+  const handleTranscript = (text: string) => {
+    // Read current textarea value instead of relying on currentMessage prop closure,
+    // which may be stale if the user typed while transcription was in-flight.
+    const current = textareaRef.current?.value ?? "";
```
akhileshrangani4: `handleTranscript` reads `textareaRef.current?.value`, but `ref={textareaRef}` is passed to `PromptInputTextarea`. If that component doesn't `forwardRef` to a native `<textarea>`, the ref is null/wrong and appended transcript gets dropped (and autosize/focus may break too). Worth verifying it forwards the ref. Safer would be a functional updater: `setCurrentMessage(prev => prev ? prev + " " + text : text)`, which reads the latest state without touching the DOM.

akhileshrangani4 (changes requested):
Really nice feature, and the code shows a lot of care: no transcript/audio in logs, `languageDetected` as a boolean to avoid leaking origin, the duration-cost tradeoff is documented, and the rate-limit intent is spelled out. Thanks for that.

That said, there are a handful of blockers before this can merge, mostly on the public `/api/transcribe` surface (it takes paid Whisper calls from unauthenticated callers, so the abuse surface matters).

**Blockers (this review):**
1. Fail-open rate limiting on the paid public endpoint (`route.ts:135`)
2. Spoofable `X-Forwarded-For` lets callers rotate rate-limit buckets (`route.ts:87`)
3. Authed surface skips the approved/banned gate everywhere else enforces (`route.ts:67`)
4. Transcription fetch has no AbortController, leaks across unmount (`VoiceInputButton.tsx:74`)
5. DB calls outside the try/catch return HTML 500 instead of the JSON error contract (`route.ts:91`)
6. DB helper imported from the tRPC router file, cross-layer leak (`route.ts:6`)
7. shareToken DB lookup actually runs before the rate-limit check, opposite of what the comment claims (`route.ts:83`)
8. `ref` on `PromptInputTextarea` may not reach a native textarea (`ChatInput.tsx:86`)

**Also: the POST handler has zero integration tests.** The helpers (`checkContentLength`, `mapProviderError`, `validateAudioBlob`) are well covered, but the handler itself, auth resolution, rate-limit wiring, the post-call duration gate, the 401/404/429/500 branches, has nothing. That is exactly where a wiring regression would slip through. Worth adding route-level tests that mock auth/rate-limit/`transcribeAudio` while fixing the above.

**Suggested fix order:** 1+2 together (they combine into a free Whisper proxy), then 3, then 4, then 5, then tests, then 6.

Leaving the smaller stuff (status codes, dead code, a couple of UI races, perf) as a separate nit pass so this one stays focused on the blockers. Full multi-reviewer report has 23 findings total.

akhileshrangani4 (commented):
Smaller stuff, none of these block. Status codes, a bit of dead code, some UI-race hardening in the recorder hook, and a couple of test gaps. Take or leave the P3s at the bottom.

apps/web/src/app/api/transcribe/helpers.ts
```diff
@@ -0,0 +1,114 @@
...
+import type { TranscriptionError } from "@teachanything/ai";
+
+/**
+ * Stable, machine-readable error codes returned by the transcription
+ * endpoint. The client uses these to map errors to localized UI rather
+ * than parsing the English `error` message.
+ */
+export type TranscribeErrorCode =
+  | "feature_disabled"
+  | "content_length_required"
```
akhileshrangani4: Nit: `content_length_required` is in the union but never emitted. `checkContentLength` returns `content_length_invalid` / `request_too_large`, and missing Content-Length is a pass-through. Dead code that could mislead client authors into handling a code that never arrives. Drop it.

apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
...
+    inferredChatbotId: chatbot.id,
+    userId: null,
+  };
+}
+
+export async function POST(
+  request: NextRequest,
+): Promise<NextResponse<SuccessResponse | ErrorResponse>> {
+  if (env.NEXT_PUBLIC_VOICE_INPUT_ENABLED === "false") {
+    return jsonError("Voice input is disabled", "feature_disabled", 404);
```
akhileshrangani4: Nit: 404 for a disabled feature reads as "endpoint doesn't exist" to health checkers and future clients. 503 (with a Retry-After) or 403 is more accurate for a kill switch. The client already keys off the `code` field so the UI is fine either way, but the status itself is part of the contract.

apps/web/src/app/api/transcribe/route.ts
```diff
@@ -0,0 +1,282 @@
...
+    if (effectiveChatbotId) {
+      try {
+        await db.insert(analytics).values({
+          chatbotId: effectiveChatbotId,
+          eventType: "voice_transcription",
+          eventData: {
+            audioBytes: validation.size,
+            durationSeconds: result.durationSeconds,
+            transcriptLength: result.text.length,
+            surface: logContext.surface as "authenticated" | "shared",
```
akhileshrangani4: Nit: `logContext.surface as "authenticated" | "shared"` casts back out of an opaque `Record<string, unknown>` bag, so the type system can't catch a wrong value. Cleaner to add a first-class `surface` field to `AuthContext` and read it directly.

apps/web/src/components/chat/messages/VoiceInputButton.tsx
```diff
@@ -0,0 +1,261 @@
...
+      setIsTranscribing(true);
+      try {
+        const form = new FormData();
+        // Hint OpenAI with a useful filename; the server validates mime
+        // independently so this is informational only.
+        const ext = blob.type.includes("mp4")
+          ? "mp4"
+          : blob.type.includes("ogg")
+            ? "ogg"
+            : "webm";
```
akhileshrangani4: Nit: this mime-to-extension fallback (mp4/ogg, else webm) is a partial duplicate of `MIME_TO_EXTENSION` in `transcription-validation.ts`, which also covers wav/mpeg/etc. They'll drift. Either export a helper from the validation module and reuse it, or just hardcode `"webm"` since the server ignores the filename hint anyway.

apps/web/src/server/routers/chatbot.ts
```diff
@@ -19,10 +19,19 @@ const allAcceptedModels = [...SUPPORTED_MODELS, ...DEPRECATED_MODELS] as [
   ...string[],
 ];
 
-async function getChatbotByIdForUser(
-  db: Parameters<
-    Parameters<typeof protectedProcedure.query>[0]
-  >[0]["ctx"]["db"],
+type ChatbotDb = Parameters<
```
akhileshrangani4: Nit (ties into the layer-leak blocker): `ChatbotDb = Parameters<Parameters<typeof protectedProcedure.query>[0]>[0]["ctx"]["db"]` couples these helpers to tRPC internals. Also `findChatbotForUser` is exported but only used internally by `getChatbotByIdForUser` in this file. When you move `findOwnedChatbotId` out to a queries module, give it a concrete Drizzle type and un-export `findChatbotForUser`.

apps/web/src/components/chat/messages/ChatInterface.tsx
```diff
@@ -29,6 +29,9 @@ interface ChatInterfaceProps {
   showFrame?: boolean;
   showSources?: boolean;
   brandingText?: React.ReactNode;
+  shareToken?: string;
+  chatbotId?: string;
+  voiceInputEnabled?: boolean;
```
akhileshrangani4, line 35: Nit: `ChatInterface` is a general layout component that now carries three voice-only props it doesn't use, just forwarding them to `ChatInput`. Consider collapsing into one `voiceConfig?: { shareToken?; chatbotId?; enabled? }` prop, or rendering `ChatInput` directly from the call sites for voice config.

apps/web/src/components/chat/messages/VoiceInputButton.tsx
```diff
@@ -0,0 +1,261 @@
...
+          } else if (code === "unauthorized") {
+            toast.error("Please sign in again to use voice input.");
+          } else {
+            toast.error(
+              "Voice input failed. Please try again or type your question.",
+            );
+          }
+          return;
+        }
+        const data = (await res.json()) as { text?: unknown };
```
akhileshrangani4, line 170: Nit: only `data.text` is read here (safe), but the success response also returns `language` and `durationSeconds` as nullable. If anything else ever consumes this endpoint, the nullable semantics aren't documented anywhere. A shared exported response type would help.

apps/web/src/hooks/useVoiceRecorder.ts
```diff
@@ -0,0 +1,291 @@
...
+      );
+      return;
+    }
+    // Synchronous lock — `status` reads aren't reliable across rapid
+    // double-clicks because setStatus is async.
+    if (startingRef.current || recorderRef.current) return;
+    startingRef.current = true;
+
+    setStatus("requesting_permission");
+    cancelledRef.current = false;
```
akhileshrangani4, line 172: Nit (race): `cancelledRef.current` is reset to false here, synchronously, before `getUserMedia` resolves. If `cancel()` fires while the permission prompt is pending, it sets the ref true, but the resolved continuation still builds a MediaRecorder and only checks `cancelledRef` later in the stop listener. Add a `cancelledRef.current === true` check right after the `getUserMedia` await: if cancelled, stop the acquired tracks and return without constructing a recorder.

apps/web/src/hooks/useVoiceRecorder.ts
```diff
@@ -0,0 +1,291 @@
...
+
+      setStatus("idle");
+      setElapsedMs(0);
+
+      if (wasCancelled) return;
+      if (blob.size === 0) {
+        emitError("no_audio", "No audio was captured. Please try again.");
+        return;
+      }
+      onComplete(blob);
```
akhileshrangani4: Nit (closure + ordering): the stop listener closes over `onComplete` from the render where `start()` ran, so a parent that swaps `onTranscript` mid-recording delivers the transcript to the old handler. You've put `onComplete` in `start`'s dep array (line 288) to compensate, but that recreates `start` on every parent render. Cleaner: store `onComplete` in a ref (same pattern as `cancelledRef`), read `ref.current(blob)` in the listener, drop it from deps. Also `cleanup()` runs before `onComplete(blob)`, so if `onComplete` throws synchronously the refs are already nulled and the error is unhandled. Wrap the call or move cleanup after it resolves.

apps/web/src/hooks/useVoiceRecorder.ts
```diff
@@ -0,0 +1,291 @@
...
+    // Recorder is now live; reentry is blocked by recorderRef from here.
+    startingRef.current = false;
+    setStatus("recording");
+    setElapsedMs(0);
+
+    tickRef.current = setInterval(() => {
+      if (startedAtRef.current !== null) {
+        setElapsedMs(Date.now() - startedAtRef.current);
+      }
+    }, 100);
```
akhileshrangani4: Nit (perf): a 100ms tick re-renders the button ~10x/sec for the whole recording (up to ~1800 renders for a 3-min clip), and the display is just MM:SS that nobody perceives below ~500ms. Bump to 500ms or 1000ms.

packages/ai/src/__tests__/transcription.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    mockFetch(
+      (async () => new Response("err", { status: 500 })) as typeof fetch,
+    );
+    const audio = new Blob([new Uint8Array([1])], { type: "audio/webm" });
+    await expect(
+      transcribeAudio({ apiKey: "k", audio, filename: "rec.webm" }),
+    ).rejects.toMatchObject({ reason: "provider_error" });
+  });
+
+  it("throws TranscriptionError with timeout on AbortError", async () => {
```
akhileshrangani4, line 112: Nit (test gap): only the AbortError path is covered. `transcription.ts` also maps a non-AbortError fetch throw to `reason: "network"`, which has no test. Also `jest.useFakeTimers()` is set up but never advanced, so the real `setTimeout -> abort` wiring isn't actually exercised (the test throws AbortError directly). Either add a `jest.advanceTimersByTime(DEFAULT_TIMEOUT_MS + 1)` test or drop the fake timers.

apps/web/src/components/chat/messages/ChatInput.tsx
```diff
@@ -69,11 +106,20 @@ export function ChatInput({
...
           <PromptInputTextarea
+            ref={textareaRef}
             placeholder="Ask me anything..."
             aria-label="Type a message"
             className="flex-1 text-foreground text-base min-h-[60px] md:min-h-[120px] scrollbar-thin"
           />
           <PromptInputActions>
+            {showVoiceInput && !isStreaming && (
+              <VoiceInputButton
+                disabled={isStreaming}
```
akhileshrangani4, line 125: Nit (P3): `disabled={isStreaming}` is redundant since the button is already behind `!isStreaming` at line 115, so it never renders while streaming. Pick one source of truth to avoid future drift.

apps/web/src/app/api/transcribe/helpers.ts
```diff
@@ -0,0 +1,114 @@
...
+        "provider_timeout",
+        504,
+      );
+    case "provider_rate_limit":
+      return jsonError(
+        "Transcription service is temporarily unavailable.",
+        "provider_unavailable",
+        503,
+      );
+    default:
```
akhileshrangani4: Nit (P3): the `mapProviderError` switch isn't exhaustive (the `network` reason falls through to the default). Behavior is fine and tested, but a `default:` with a `const _exhaustive: never = err.reason` would make a future added reason a compile error instead of silent contract drift.

akhileshrangani4 (approved):
All blockers from the last round are addressed, and nearly all the nits too. Went through the new code fresh as well and it holds up. Approving.

Verified against the previous review:
- fail-closed rate limiting on the public surface via `requireRateLimit`, plus the new per-shareToken global cap so IP rotation can't unbound the spend. Both tested.
- trusted IP derivation (`x-real-ip`, rightmost XFF) in `get-client-ip.ts`, with a test that specifically pins the spoofed-leftmost case
- approval/ban gate now mirrors `protectedProcedure` (admin bypass included), tested
- AbortController + mounted ref on the client fetch, plus the 100s client timeout sitting above the 90s provider cap
- DB failures return the JSON error contract on both the shareToken lookup and the ownership lookup, tested
- `findOwnedChatbotId` moved to `server/queries/chatbot.ts` with a concrete Drizzle type, router re-exports for compatibility
- rate limit genuinely runs before the shareToken lookup now, and there's a test pinning that ordering so it can't regress

The Jul 3 hardening is good work too: the real-timer timeout test, mapping an abort during the body read to `timeout`, keeping the mic button mounted-but-disabled during streaming so a routine send can't destroy an in-flight transcript, and releasing the mic when the component unmounts while `getUserMedia` is pending.

Two tiny take-it-or-leave-it notes, neither blocking:
- `useVoiceRecorder.stop()`: if `recorder.stop()` throws, status stays `"stopping"` and the button spins until remount. A `setStatus("idle")` in that catch would cover it.
- `ChatInterface` still forwards the three voice props individually rather than a single config object. Fine as-is, just noting the earlier nit was consciously declined, which works for me.

I'll resolve the merge conflicts with main on this branch and push directly, so nothing needed on your end there. Thanks for the thorough follow-through on this one.

PR #313, opened 2026-06-06
feat(chat): structured study modes (quiz, flashcards, test, mind map, matching) with confirmation gate

gum1x:
## Summary

Adds **structured study modes** to the chatbot: when a student asks for a study tool, the bot returns validated JSON the client renders as an interactive widget instead of plain prose.

Modes included:
- **Quiz** — multiple-choice, scored
- **Flashcards** — flip-through study deck
- **Test** — longer graded test with a timer, now supporting **free-response (open-ended) questions** graded by the AI as a chat follow-up
- **Mind map** — collapsible hierarchical concept map
- **Matching pairs** — Duolingo-style two-column matching game with first-try accuracy

It also replaces silent keyword triggering with a **Yes/No confirmation step**, so the bot never generates the wrong format ("don't make flashcards" no longer makes flashcards) and no longer misses casual phrasings ("i want u to make flashcards…").

## How it works

Each mode is a `StructuredMode` descriptor (`detect`, `instruction`, `schema`, `summarize`, `fallback`) driven by an ordered registry. The streaming pipeline:

1. **Eager-detect** a study-tool request → emit an ephemeral `confirm` event (no LLM call) → render a Yes/No card.
2. **Yes** → send a canonical trigger phrase (guaranteed to pass the strict detector) → the model returns strict JSON, which the server buffers (capped to guard runaway streams), validates, and emits as a single structured event.
3. **No** → answer the original message as normal chat.
4. The client re-validates the payload and renders the matching widget. Payloads persist in the `messages` JSONB metadata, typed by a single `StructuredModeId` source of truth.

## Notable details

- **Back-compat**: legacy persisted tests (multiple-choice questions without a `type` field) still parse — the schema defaults `type` to `"multiple_choice"`.
- **Confirmation gate** handles intent that regex can't: leading negations and "what is…" questions are excluded; the canonical-trigger round-trip is unit-tested for every mode.
- Hardening: structured-buffer size cap, matching-pair uniqueness validation, duplicate-option React keys, and failure toasts.

## Testing

- `npm run check-types` ✅
- `npm run lint` ✅
- `npm run test` ✅ — **411 tests pass**, including new coverage for each mode's schema/triggers, eager detection + negation guards, canonical-trigger round-trips, and grading helpers.

Verified to build and pass off a clean `main` (no dependency on unrelated in-flight work).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/web/src/__tests__/lib/flashcards.test.ts
+++ b/apps/web/src/__tests__/lib/flashcards.test.ts
@@ -1,0 +1,92 @@
+import { describe, it, expect } from "@jest/globals";
+import { isFlashcardRequest, flashcardsSchema } from "@/lib/flashcards";
+
+describe("isFlashcardRequest", () => {
+  it("matches common flashcard phrasings", () => {
+    const positives = [
+      "make flashcards",
+      "Make flashcards",
+      "  MAKE FLASHCARDS  ",
+      "flashcard me",
+      "flash card me",
+      "study cards",
+      "give me flashcards on python",
+      "make me flashcards on the cell cycle",
+      "create flashcards on photosynthesis",
+      "can you make flashcards",
+    ];
+    for (const message of positives) {
+      expect(isFlashcardRequest(message)).toBe(true);
+    }
+  });
+
+  it("does not match unrelated messages", () => {
+    const negatives = [
+      "what is a flashcard",
+      "I made flashcards yesterday",
+      "tell me about the syllabus",
+      "explain recursion",
+      "",
+      "thanks!",
+    ];
+    for (const message of negatives) {
+      expect(isFlashcardRequest(message)).toBe(false);
+    }
+  });
+});
+
+describe("flashcardsSchema", () => {
+  const validDeck = {
+    deck_title: "Python Basics Deck",
+    cards: [
+      {
+        front: "What keyword defines a function?",
+        back: "The 'def' keyword defines functions in Python.",
+      },
+      {
+        front: "What does print() do?",
+        back: "It writes output to standard output.",
+      },
+    ],
+  };
+
+  it("parses a well-formed deck", () => {
+    const result = flashcardsSchema.safeParse(validDeck);
+    expect(result.success).toBe(true);
+  });
+
+  it("rejects a deck with no cards", () => {
+    const result = flashcardsSchema.safeParse({
+      deck_title: "Empty",
+      cards: [],
+    });
+    expect(result.success).toBe(false);
+  });
+
+  it("rejects a deck with more than ten cards", () => {
+    const result = flashcardsSchema.safeParse({
+      deck_title: "Too many cards",
+      cards: Array.from({ length: 11 }, (_, i) => ({
+        front: `front ${i}`,
+        back: `back ${i}`,
+      })),
+    });
+    expect(result.success).toBe(false);
+  });
+
+  it("rejects a card missing its front", () => {
+    const result = flashcardsSchema.safeParse({
+      deck_title: "Missing front",
+      cards: [{ front: "", back: "an answer" }],
+    });
+    expect(result.success).toBe(false);
+  });
+
+  it("rejects a card missing its back", () => {
+    const result = flashcardsSchema.safeParse({
+      deck_title: "Missing back",
+      cards: [{ front: "a question", back: "" }],
+    });
+    expect(result.success).toBe(false);
+  });
+});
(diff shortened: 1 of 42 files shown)


akhileshrangani4 (changes requested):
Really nice work and it clearly works. But I don't think we should merge this as-is, and it's a bigger-picture thing than line nits: this is a hand-rolled reimplementation of generative UI that `ai@6` already gives us natively, and I don't want us owning and maintaining all this custom surface (a mode registry, a custom stream-event protocol, regex intent detection, a flat-union-plus-casts message type, a hand-written buffer/parse/fallback path) when the SDK does it for us. So flagging this as blocking, let's talk before it lands.

Quick walkthrough of the native approach, since it's not obvious if you haven't used AI SDK tools. The idea: register each study mode as a *tool*. The model calls the tool and fills in the structured payload as the tool input (your existing Zod schema validates it), and you render that payload as the widget. No regex detection, no `JSON.parse`, no buffering, no fallback string.

**1. Mode = a tool** (server). Your `quizSchema` becomes the tool's `inputSchema`:
```ts
import { tool } from 'ai';
import { quizSchema } from '@/lib/modes/quiz';

const tools = {
  showQuiz: tool({
    description: 'Generate a quiz when the student asks for one',
    inputSchema: quizSchema,        // model fills this in = the quiz
    needsApproval: true,            // <-- this is the confirm gate
    execute: async (quiz) => quiz,  // no-op; payload is the input
  }),
  // showFlashcards, showTest, showMindMap, showMatching...
};

const result = streamText({ model, messages, tools });
return result.toUIMessageStreamResponse();
```
That deletes `detection.ts` (the model decides when to call `showQuiz`) and the whole buffer -> `JSON.parse` -> `schema.parse` -> `fallbackMessage` block in `chat.ts`.

**2. The confirm card is `needsApproval: true`.** The SDK pauses and emits a part with `state: 'approval-requested'` instead of running the tool. Client:
```tsx
import { useChat } from '@ai-sdk/react';
import { lastAssistantMessageIsCompleteWithApprovalResponses } from 'ai';

const { messages, sendMessage, addToolApprovalResponse } = useChat({
  sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithApprovalResponses,
});
```
That replaces `ConfirmModeCard` + the `confirm` event + `confirmYes`/`confirmNo`.

**3. Render by walking `message.parts`** (replaces the if/else + casts in `ChatMessage.tsx`):
```tsx
message.parts.map((part) => {
  if (part.type === 'tool-showQuiz') {
    switch (part.state) {
      case 'approval-requested':
        return (
          <div>
            Make a quiz on this?
            <button onClick={() => addToolApprovalResponse({ id: part.approval.id, approved: true })}>Yes</button>
            <button onClick={() => addToolApprovalResponse({ id: part.approval.id, approved: false })}>No</button>
          </div>
        );
      case 'output-available':
        return <QuizMessage quiz={part.input} />;  // typed, no `as Quiz`
    }
  }
});
```
`parts` is a discriminated union, so `part.input` is typed as `Quiz` inside that branch. The `messageType?`/`structured?` fields and every `as Quiz`/`as Test` cast go away.

Refs: [human-in-the-loop cookbook](https://ai-sdk.dev/cookbook/next/human-in-the-loop), [chatbot tool usage](https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-tool-usage), [AI SDK 6](https://vercel.com/blog/ai-sdk-6).

**Two things I checked in our codebase so we go in eyes-open, not blockers to the approach but real scope:**
- This isn't structured-modes-only. Our chat runs over a tRPC subscription (`sendSharedMessageStream`/`sendMessageStream`) with our own `handleStreamData` event union across `useChat.ts`, `useChatbot.ts`, `useChatState.ts`, and `@ai-sdk/react` isn't installed yet. The client bits above assume the AI SDK chat transport, so moving to it touches the whole chat path, not just the new code. That's exactly why I'd rather do it now than add more custom code on top.
- One behavior to preserve: our current gate confirms *before* generation (the eager-detect path early-returns in `chat.ts` before `streamText`, so we don't spend tokens on a quiz the user didn't want). `needsApproval` pauses *after* the model produced the tool input, so to keep the cheap pre-generation confirm we'd use a lightweight intent tool + approve, then generate.

Happy to pair on the migration if useful.

akhileshrangani4:
Closing this as we are implementing according to the new plan.

PR #340, opened 2026-06-11
feat(admin): add registration email templates and admin email actions

### LINKED ISSUE #136: feat(admin): Add automated email templates for incomplete user registrations
## Context

The platform uses an admin approval workflow for user registration. Admins review signups and approve or reject them. However, **people frequently sign up with incomplete or incorrect information**, and there's currently no way for admins to contact them directly from the dashboard.

Right now, an admin would have to manually copy the user's email, open their email client, and write a message from scratch. This should be a one-click action with pre-defined email templates.

## Relevant Code

| File | Purpose |
|------|---------|
| `apps/web/src/server/routers/admin.ts` | Admin router — user management endpoints (`getUsers`, `updateUserStatus`, etc.) |
| `apps/web/src/lib/email.ts` | Email service — all emails routed through QStash via `queueEmail()` |
| `apps/web/src/lib/qstash.ts` | QStash client — `publishEmailJob()` for async email delivery |
| `apps/web/src/app/api/jobs/send-email/route.ts` | QStash job handler — sends via Resend with idempotency and delivery tracking |
| `apps/web/src/app/api/webhooks/resend/route.ts` | Resend webhook handler — tracks delivery status (sent → delivered → bounced) |
| `packages/db/src/schema.ts` | `emailDeliveries` table for tracking, `emailTypeEnum` for allowed types |
| `apps/web/src/components/emails/` | React Email components (9 templates — follow this pattern) |

### Current email sending pattern:

All emails go through QStash for reliable async delivery with tracking. New emails **must** follow this pattern:

```typescript
// 1. Create a React Email component in components/emails/
// 2. Add a new value to emailTypeEnum in schema.ts
// 3. Add a wrapper function in email.ts that calls queueEmail()

async function queueEmail(params: {
  emailType: EmailType;       // Must be in emailTypeEnum
  to: string | string[];
  subject: string;
  reactComponent: React.ReactElement;
  replyTo?: string;
}): Promise<{ deliveryId: string }>
```

This creates a delivery tracking record, publishes to QStash, and the `/api/jobs/send-email` handler sends via Resend with idempotency keys.

### Existing email types (for reference):
`admin_notification`, `approval`, `rejection`, `promote_admin`, `demote_admin`, `account_disabled`, `account_enabled`, `account_deleted`, `password_reset`

### 1. New email templates (React Email components)
Create new React Email components in `components/emails/`:
- **`RequestMoreInfo.tsx`**: "We need additional information to complete your registration..."
- **`IncorrectInfo.tsx`**: "Some information in your registration appears incorrect..."
- **`GenericAdminMessage.tsx`**: A template that accepts a custom message body from the admin

Each template should:
- Follow the existing React Email component patterns
- Include the user's name (auto-populated)
- Include a reply-to address or link back to the platform
- Be consistent with the existing email styling

### 2. Schema update
- Add new values to `emailTypeEnum`: `request_more_info`, `incorrect_info`, `generic_admin_message`
- Run `npm run db:generate` for the migration

### 3. Backend: Email sending endpoint
- Add wrapper functions in `email.ts` following the existing pattern (e.g., `sendRequestMoreInfoEmail`, `sendIncorrectInfoEmail`, `sendGenericAdminEmail`)
- Add a new tRPC procedure in the admin router:
  - `sendRegistrationEmail`: Admin-only procedure that sends a templated email to a user
  - Input: `{ userId, templateId: "request_more_info" | "incorrect_info" | "generic_admin_message", customMessage?: string }`
  - Validates that the user exists
  - Calls the appropriate email wrapper function

### 4. Frontend: Email action in admin dashboard
- Add a **"Send Email"** button or dropdown next to each user in the admin user list
- Clicking it opens a modal/dialog with:
  - Template selection (dropdown or radio buttons)
  - Preview of the email that will be sent
  - Optional custom message field (for the generic template)
  - Send button
- Show success/error feedback after sending
- Leverage existing Shadcn UI components (Dialog, Select, Textarea, Button)

## Acceptance Criteria

- [ ] Admins can send pre-defined template emails to users from the admin dashboard
- [ ] At least 3 templates are available (request info, incorrect info, generic)
- [ ] Email preview is shown before sending
- [ ] Generic template allows custom message input
- [ ] User's name is auto-populated in the email
- [ ] Emails go through the existing QStash pipeline (delivery tracking, idempotency, retries)
- [ ] New email types are added to `emailTypeEnum` with a migration
- [ ] Only admin users can access the send email functionality
- [ ] Success/error feedback is shown after sending
- [ ] `npm run check-types` and `npm run lint` pass

## Technical Notes

- **Follow existing patterns exactly**: Look at how `sendApprovalEmail` works end-to-end — React Email component → `queueEmail()` → QStash → Resend. New emails must follow this same flow.
- **Rate limiting**: Consider adding rate limiting to prevent accidental spam (e.g., max 5 template emails per user per day).
- **No new dependencies needed** — Resend, React Email, and QStash are already configured.

## Contributing Guidelines

- Follow the coding standards in `AGENTS.md` (strict TypeScript, no `any`, kebab-case files, etc.)
- PR title must use conventional commit format: `feat(admin): add email templates for user registrations`
- Run `npm run lint && npm run check-types` before submitting
- Run `npm run db:generate` for the schema migration
- Link this issue in your PR description

M-Rakshitha:
## Summary

Implements automated registration email workflows for admins reviewing pending user registrations.

Closes #136

Key changes:
- Added three new registration email templates
- Added admin email actions in the pending users dashboard
- Added backend support for sending templated registration emails
- Added email delivery tracking support through the existing QStash/Resend pipeline
- Added validation and test coverage for the new functionality

## Changes

Added/Modified/Removed:
- Added admin email template support for pending user approvals.
- Added Send Email action for each pending user in `PendingUsersTab`.
- Added email preview and template selection UI in the send-email dialog.
- Added tests covering email wrapper exports and the admin `sendRegistrationEmail` mutation.
- Added/refined handler-level logic for `sendRegistrationEmail` to make it testable.
- Reverted the temporary extra pending seed user so the demo data stays unchanged.

Architectural decision:
- Separated `sendRegistrationEmail` into a standalone handler:
makes business logic testable without router/auth import side effects.
- Kept the admin router thin and wired it to the shared handler:
preserves clear API/resolver structure and avoids duplicate validation.
- Used the existing qstash / `email` wrapper flow:
keeps email delivery consistent with the app’s current background-job pipeline.
- Rendered the send-email action per pending row:
ensures admins can message any new signup directly from the pending approvals list.

## Library Choices (if applicable)

- No new dependencies added

## Screenshots (if visual changes)

Before:
Screenshot 2026-06-11 at 12 21 30 PM

After:
Send Email Button added:
Screenshot 2026-06-11 at 12 11 44 PM

Clicking the Send Email button:
Screenshot 2026-06-11 at 12 40 40 PM

## Test Plan

- [x] How to test the changes
  - Log in as an admin user
  - Navigate to the Pending Users page
  - Click the new **Send Email** action for a pending user
  - Select each available template:
    - Request More Information
    - Incorrect Information
    - Generic Admin Message
  - Send the email and verify success feedback is displayed

- [x] Edge cases considered
  - Invalid user ID
  - Invalid template selection
  - Missing required input values
  - Generic template without a custom message
  - Email rate limiting / send count enforcement
  - Non-admin access attempts

- [x] What was manually verified
  - Send Email button appears for pending users
  - Email dialog opens correctly
  - All three templates can be selected
  - Email preview displays correctly
  - Success feedback is shown after sending
  - Registration email workflow executes successfully in development mode
  - Validation errors are handled correctly

## Checklist

- [x] PR title follows Conventional Commit format
- [x] Linked to an issue (or context provided)
- [x] `npm run lint` passes
- [x] `npm run check-types` passes
- [x] `npm run test` passes
- [x] No `console.log` statements
- [x] Zod validation on all new tRPC inputs
- [x] Ownership checks on protected resources
- [x] Screenshots attached for visual changes
- [x] New dependencies justified in PR description (No new dependencies)
- [x] Tests added for new utility/validation logic
- [x] Database migrations generated (`npm run db:generate`) if schema changed
- [x] I understand and can stand behind every change in this PR


M-Rakshitha:
Hi Akhilesh, 

 I noticed there is now a new merge conflict due to more recent changes on main, and I'm happy to update the branch again if needed. I saw the Codecov report as well, so if you'd like me to add additional test coverage or make any other changes before this is reviewed, I'd be happy to do so. Whenever you have a chance, I'd appreciate any feedback. Thank you!

akhileshrangani4 (changes requested):
Hey @M-Rakshitha, sorry this sat so long. I went through the whole thing properly just now.

Short version: the feature is still wanted and nothing has shipped in the meantime that overlaps it, so none of your work is dead. But `main` moved a lot in June with the architecture cleanup in #458, and the branch conflicts now. Could you rebase onto latest `main`? Below is what changed underneath you, so you're not rebasing blind, plus the review so you can fold both into one pass instead of two.

## What moved under the branch

The three real conflicts:

- `server/routers/admin.ts` was deleted. It's now `server/routers/admin/index.ts` (just a registration map) plus one file per concern in `admin/procedures/`. Your 13 line addition becomes `admin/procedures/registration-email.ts` plus one line in the index.
- `lib/email.ts` was split into `server/email/delivery.ts` (`queueEmail`), `server/email/notifications.ts` (the `send*Email` wrappers) and a barrel at `server/email/index.ts`. Everything imports from `@/server/email` now. Good news is `queueEmail` itself is unchanged, same signature, so your three wrappers drop into `notifications.ts` nearly as is.
- `PendingUsersTab.tsx` was split too. The entry point is query wiring only now, dialogs live in `pending-users-tab/pending-users-dialogs.tsx`, actions in `pending-users-tab/use-pending-users-actions.ts`.

These four merge clean but will still break the build, which is the sneaky part:

- `admin-send-registration-email.ts` imports `@/lib/email`, which no longer exists
- `__tests__/server/admin-email-mutation.test.ts` mocks `@/lib/email` and `@/lib/logger`
- `__tests__/lib/email-wrappers.test.ts` imports `@/lib/email`
- `server/routers/admin.ts` gets left in the tree as your version, so delete it rather than resolving it

Two smaller things: migration slot 0019 is still free so your number works out, and `requireExistingUser` now exists in `admin/procedures/helpers.ts`, which replaces the lookup block you hand rolled.

I'm holding off on inline comments since a rebase force push would orphan them all. Happy to do a proper line by line pass once it's rebased.

## Blockers

**1. The migration was hand written rather than generated.** `packages/db/drizzle/0019_true_mole_man.sql` edits `meta/0014_snapshot.json`, which is a historical snapshot that has to stay frozen, and adds no `0019_snapshot.json`. The journal timestamp for 0019 is also older than 0017's, so the next `db:generate` will diff against a state that never existed. Please delete the sql file, revert both `meta/` files, keep only the enum change in `schema.ts`, and run `npm run db:generate` after the rebase. Slot 0019 is free so you'll get the same number back.

**2. `seed.ts` throws.** The new early return block interpolates `demoPwHash` at lines 200 and 211, but the `const` is declared at 222. Any run where the professor exists and the student doesn't hits a TDZ `ReferenceError`. Honestly I'd just drop the seed change from this PR. It's a fair improvement but it's unrelated to admin emails and it's costing you a review round.

**3. The templates say `Teach Anything™`.** We switched everything to `®` in #430 after the mark was registered, and all nine existing templates use it. Four spots across the three new templates need changing.

## Worth fixing

**4. The procedure doesn't check that the user is pending.** It looks up by id and sends, so an admin can send 1000 characters of free text to any account in the system, including other admins, under our from address with a subject that reads like an official account notice. The UI only shows the button on the pending tab but the endpoint is the boundary, not the UI. Please gate on `status === "pending"` and throw `BAD_REQUEST` otherwise.

**5. Rate limiting.** Two things here. `server/rate-limit.ts` already has `adminActionRateLimit` and AGENTS.md asks for those to be used, so the hand rolled version isn't needed. More importantly the count over `emailDeliveries` has no `deliveryStatus` filter, and `queueEmail` writes a `failed` row when the QStash publish throws. Five QStash hiccups and that recipient is locked out for 24 hours with nothing the admin can do about it. It's also keyed per recipient rather than per admin, so it isn't really an abuse control.

**6. Nothing records who sent what.** Subject is a fixed string, so a free text message sent under the platform's name leaves no trace of its content or its author. `emailDeliveries` already has a `metadata` jsonb column, so threading `{ sentByAdminId, templateId }` through would cover it. Your tests already mock the logger, which suggests you were heading here anyway.

**7. The dialog preview retypes the email copy in JSX.** Two sources of truth, and they've already drifted: the preview says "reply to the support email included with this message" while the template says "reply to this email or reach out to our support team". Either render the preview from the same components or preview only the subject and the custom message.

**8. The dialog needs to go in its own file.** `PendingUsersTab.tsx` is 270 lines now and your +277 would push it past 500, and that file was deliberately split. New `pending-users-tab/send-email-dialog.tsx`, kebab case to match that directory, mutation alongside approve and reject in `use-pending-users-actions.ts`.

**9. Two test issues.** `admin-email-validation.test.ts` copy pastes the zod schema instead of importing the exported one, so it tests a copy and would keep passing if the real schema changed. And `email-wrappers.test.ts` only asserts `typeof x === "function"` on each export, which the type checker already does. That's why codecov flagged the file at 26 percent. Something like rendering `RequestMoreInfo` and asserting the login url is present, or asserting that a `<script>` in `customMessage` comes out escaped, would be real coverage.

**10. Global `console.warn` mock in `jest.setup.ts`.** That silences warnings for every test in `apps/web` to quiet one path, and `queueEmail`'s dev fallback uses `console.warn` to tell you an email wasn't actually sent. Please scope it to the file that needs it.

## Nits

- `validInput.customMessage!.trim()`, the non null assertion. A discriminated union on `templateId` would let the compiler know, and would make `customMessage` an error on the two templates that ignore it.
- The handler re-parses the input that tRPC already parsed. Once the schema is exported for the test, the "for testability" comment doesn't apply.
- `RequestMoreInfo` and `IncorrectInfo` both say "reply to this email" in the body and "do not reply directly to this email" in the footer. These three are the only templates that set `replyTo`, so replies genuinely do work and the boilerplate footer is the wrong one.
- `GenericAdminMessage` renders the custom message in a single `<p>`, so newlines collapse and anyone writing two paragraphs gets a wall of text. `whiteSpace: "pre-wrap"` fixes it. The escaping itself is fine, React handles that.
- The diff adds `replyTo` to the existing `sendRejectionEmail`. Reasonable change, just call it out in the description or split it.

One thing I want to be clear about since it's a long list: the structure is right. Pulling the handler into its own file was the correct instinct and it's actually more idiomatic now than when you wrote it, the templates match how the other nine are built, and using `unstable_mockModule` matches what the rest of the suite does even though AGENTS.md still shows the older pattern. Most of the rebase is mechanical path fixing rather than rewriting.

Ping me when it's rebased and I'll turn it around quickly this time.

packages/db/drizzle/0019_true_mole_man.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TYPE "public"."email_type" ADD VALUE 'request_more_info' BEFORE 'promote_admin';--> statement-breakpoint
```
akhileshrangani4, line 1: Blocker, and the fix is mostly deletion.

This was hand written rather than generated, and the diff shows it three ways: it edits `meta/0014_snapshot.json`, which is a historical snapshot that has to stay frozen, it adds no `meta/0019_snapshot.json` (main has 0000 through 0018 and nothing else), and the `_journal.json` entry for 0019 has an older timestamp than 0017's. So the next `db:generate` diffs against a state that never existed.

What I'd do: delete this file, revert both files under `drizzle/meta/`, keep only the enum change in `packages/db/src/schema.ts`, then run `npm run db:generate` after the rebase and commit whatever it emits. Slot 0019 is still free on main so you'll get the same number back.

The SQL itself is fine by the way. `ALTER TYPE ... ADD VALUE` inside drizzle's transaction is okay on PG 12+ as long as the value isn't used in the same transaction, and Supabase is on 15.

packages/db/scripts/seed.ts
```diff
@@ -175,11 +175,44 @@ async function seed() {
...
+
+          console.log("  Restored student@demo.edu to pending approval\n");
+        } else {
+          const [pending] = await tx`
+            INSERT INTO "user" (id, email, name, email_verified, status, role, department, institutional_affiliation, created_at, updated_at)
+            VALUES (gen_random_uuid()::text, 'student@demo.edu', 'Alex Johnson', true, 'pending', 'user', 'Mathematics', 'Demo University', NOW(), NOW())
+            RETURNING id`;
+          await tx`
+            INSERT INTO "account" (id, user_id, account_id, provider_id, password, created_at, updated_at)
+            VALUES (gen_random_uuid()::text, ${pending!.id}, 'student@demo.edu', 'credential', ${demoPwHash}, NOW(), NOW())`;
```
akhileshrangani4, line 211: Blocker: this throws. `demoPwHash` is interpolated here and at line 200, but the `const demoPwHash = await bcrypt.hash(...)` is declared at 222. Any run where `professor@demo.edu` exists and `student@demo.edu` doesn't reaches this line and dies with `Cannot access 'demoPwHash' before initialization`.

My suggestion is to drop the seed change from this PR entirely. Restoring a pending demo student is a fair improvement but it's unrelated to admin emails, and shipping it broken costs you a review round for something that isn't the feature. If you'd rather keep it, hoisting the two `bcrypt.hash` calls above the early return block fixes it.

apps/web/src/components/emails/RequestMoreInfo.tsx
```diff
@@ -0,0 +1,126 @@
...
+          cellPadding="0"
+          cellSpacing="0"
+          border={0}
+        >
+          <tr>
+            <td style={cell}>
+              <p style={text}>Hi {userName},</p>
+
+              <p style={text}>
+                Thank you for registering with Teach Anything™. We need a bit
```
akhileshrangani4, line 34: Blocker, small one. Should be `Teach Anything®`. We switched the whole codebase in #430 after the mark was registered with the USPTO, and all nine existing templates use `®`.

Four spots total: this line, `RequestMoreInfo.tsx:59`, `IncorrectInfo.tsx:58` and `GenericAdminMessage.tsx:47`.

apps/web/src/server/routers/admin-send-registration-email.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  // Find the user
+  const [foundUser] = await ctx.db
+    .select({ email: user.email, name: user.name })
+    .from(user)
+    .where(eq(user.id, validInput.userId))
+    .limit(1);
+
+  if (!foundUser || !foundUser.email) {
+    throw new TRPCError({ code: "NOT_FOUND", message: "User not found" });
```
akhileshrangani4, line 62: This is the one I'd fix first.

The lookup finds the user by id and the handler sends, but nothing checks that they're actually pending. The dialog only appears on the pending tab, but the tRPC procedure is the security boundary, not the UI. As written, an admin (or a stolen admin session) can send 1000 characters of free text to any account in the system, including other admins, from our address, under a subject that reads like an official account notice. That's a usable phishing primitive and it's the kind of thing you want closed before it ships rather than after.

Two changes:

```ts
const existingUser = await requireExistingUser(ctx, validInput.userId);
if (existingUser.status !== "pending") {
  throw new TRPCError({ code: "BAD_REQUEST", message: "..." });
}
```

`requireExistingUser` is new since you branched, it's in `admin/procedures/helpers.ts` and does exactly the lookup-and-throw block you've written here. If sending to approved users is wanted later that should be its own procedure with its own name.

apps/web/src/server/routers/admin-send-registration-email.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  // Check 24-hour email count
+  const since = new Date(Date.now() - 24 * 60 * 60 * 1000);
+  const [countResult] = await ctx.db
+    .select({ count: sql<number>`count(*)` })
+    .from(emailDeliveries)
```
akhileshrangani4, line 69: Two problems with the rate limiting, and the second one will bite a real admin.

First, `server/rate-limit.ts` already has `adminActionRateLimit` (20/min per admin) and AGENTS.md section 11 asks for those to be used rather than hand rolled.

Second, and this is the real one: this count has no `deliveryStatus` filter, and `queueEmail` writes a row with `deliveryStatus: "failed"` when the QStash publish throws. So five QStash hiccups against one recipient and that recipient is locked out for 24 hours with a `TOO_MANY_REQUESTS` nobody can clear. Failed sends counting against the cap is backwards.

It's also keyed per recipient rather than per admin, so it isn't really an abuse control, and count-then-send isn't atomic so two admins can race past 5 anyway.

I'd add `await adminActionRateLimit.limit(ctx.session.user.id)` as the actual control. If you want to keep the per recipient cap as a courtesy limit, filter it to `deliveryStatus in ("queued", "sent", "delivered")`.

apps/web/src/server/routers/admin-send-registration-email.ts
```diff
@@ -0,0 +1,116 @@
...
+    .where(
+      and(
+        eq(emailDeliveries.recipientEmail, foundUser.email),
+        gte(emailDeliveries.createdAt, since),
+        or(
+          eq(emailDeliveries.emailType, "request_more_info"),
+          eq(emailDeliveries.emailType, "incorrect_info"),
+          eq(emailDeliveries.emailType, "generic_admin_message"),
```
akhileshrangani4, line 77: Related to the above: these three types are all that gets recorded, and the subject stored alongside them is a fixed string. So a free text message, typed by a human and sent under the platform's name, leaves no trace of what it said or who sent it. That's the one thing you'd want in the log six months later when someone asks what we emailed them.

`emailDeliveries` already has a `metadata` jsonb column, so threading `{ sentByAdminId, templateId }` through `queueEmail` into it would cover it, plus a `logInfo` in the handler. Your tests already mock `@/lib/logger`, which makes me think you were heading this way anyway.

apps/web/src/server/routers/admin-send-registration-email.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Handler implementation for sendRegistrationEmail mutation.
+ * Tests this function directly to avoid ESM issues with tRPC/auth imports.
+ */
+export async function sendRegistrationEmail(
+  ctx: { db: typeof DbType },
+  input: SendRegistrationEmailInput,
+): Promise<{ success: true }> {
+  // Validate input (normally done by tRPC Zod schema, but we validate here for testability)
+  const validInput = sendRegistrationEmailInputSchema.parse(input);
```
akhileshrangani4, line 52: Nit: tRPC already parsed this through `.input()`, so this is a second parse of the same object. Harmless, but once the schema is exported so the validation test can import it (see my comment on that file), the "we validate here for testability" reasoning goes away. Type the param as `SendRegistrationEmailInput` and use `input` directly.

apps/web/src/server/routers/admin-send-registration-email.ts
```diff
@@ -0,0 +1,116 @@
...
+      await sendIncorrectInfoEmail({
+        email: foundUser.email,
+        name,
+      });
+      break;
+    case "generic_admin_message":
+      await sendGenericAdminEmail({
+        email: foundUser.email,
+        name,
+        customMessage: validInput.customMessage!.trim(),
```
akhileshrangani4, line 110: Nit: the non-null assertion. The `.refine()` does guarantee it, but the type doesn't, and AGENTS.md section 6 steers away from these.

A discriminated union on `templateId` would let the compiler know it's there, and it'd go further: it would make `customMessage` a type error on the two templates that ignore it, which is currently only prevented by everyone remembering. Section 6 calls these out specifically for mutually exclusive states, which is what the three templates are.

apps/web/src/components/emails/GenericAdminMessage.tsx
```diff
@@ -0,0 +1,114 @@
...
+          style={wrapper}
+          cellPadding="0"
+          cellSpacing="0"
+          border={0}
+        >
+          <tr>
+            <td style={cell}>
+              <p style={text}>Hi {userName},</p>
+
+              <p style={text}>{customMessage}</p>
```
akhileshrangani4, line 33: Nit: this renders up to 1000 characters of admin typed text in a single `<p>`, so newlines collapse. Any admin who writes two paragraphs gets one wall of text. A dedicated style object with `whiteSpace: "pre-wrap"` fixes it.

The escaping here is correct, to be clear. React treats it as a text child so a `<script>` in the message comes out inert. That part I checked.

akhileshrangani4 (commented):
@M-Rakshitha following up on my own review, I've put everything from it inline now so you can see exactly which line each point lands on. Same findings, nothing new, just anchored where they belong.

The three blockers are on `0019_true_mole_man.sql`, `seed.ts` and the templates. The one I'd read first is on the handler's user lookup.

apps/web/src/__tests__/server/admin-email-validation.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, it, expect } from "@jest/globals";
+import { z } from "zod";
+
+// Recreate the input schema used by the admin.sendRegistrationEmail procedure
+// so we can unit-test the validation rules without invoking the router.
+const sendRegistrationEmailInput = z
```
akhileshrangani4, line 6: This re-declares the schema instead of importing it, so the file is testing a copy. Change the real schema in the handler and every case here still passes, which is the opposite of what you want from a validation test.

Export `sendRegistrationEmailInputSchema` from the procedure and import it here. It's just zod, no runtime deps, so there's no ESM reason to duplicate it.

apps/web/src/__tests__/lib/email-wrappers.test.ts
```diff
@@ -0,0 +1,58 @@
...
+process.env.NEXT_PUBLIC_APP_URL = "http://localhost:3000";
+process.env.ADMIN_EMAILS = "admin@example.com";
+process.env.NEXT_PUBLIC_CONTACT_EMAIL = "support@example.edu";
+process.env.RESEND_FROM_EMAIL = "noreply@example.edu";
+process.env.QSTASH_TOKEN = "test-token";
+
+describe("email wrappers", () => {
+  it("exports sendRequestMoreInfoEmail function", async () => {
+    const emailModule = await import("@/lib/email");
+    expect(typeof emailModule.sendRequestMoreInfoEmail).toBe("function");
```
akhileshrangani4, line 22: Every case in this file asserts `typeof x === "function"` or checks `.name`, which the type checker already guarantees. It's why codecov flagged the module at 26 percent: the file gets imported, not exercised.

I'd delete it and test the rendered output instead, which is the coverage that's actually missing. Something like: `render(RequestMoreInfo({...}))` contains the login URL and the support mailto, and `GenericAdminMessage` escapes a `<script>` and an `&` passed in `customMessage`. That would catch a real break.

apps/web/test/jest.setup.ts
```diff
@@ -3,6 +3,9 @@ import { jest } from "@jest/globals";
 import { TextEncoder, TextDecoder } from "util";
 import { TransformStream, ReadableStream, WritableStream } from "stream/web";
 
+// Suppress noisy environment warnings during tests
+jest.spyOn(console, "warn").mockImplementation(() => {});
```
akhileshrangani4, line 7: This silences `console.warn` for every test in `apps/web`, permanently, to quiet one noisy path.

Worth knowing what it hides: `queueEmail`'s dev fallback uses `console.warn` to tell you an email wasn't actually sent. That's exactly the warning someone will need one day.

Please revert this and scope the suppression to the test file that needs it, in a `beforeEach`/`afterEach` pair that restores it.

apps/web/src/components/admin/tabs/PendingUsersTab.tsx
```diff
@@ -180,6 +238,81 @@ export function PendingUsersTab() {
...
+    setSendEmailDialog({
+      isOpen: false,
+      userId: null,
+      userName: null,
+      userEmail: null,
+      templateId: "request_more_info",
+      customMessage: "",
+    });
+
+  const getPreviewBody = () => {
```
akhileshrangani4, line 266: Two things here, and both get easier after the rebase.

**The preview retypes the email copy.** `getPreviewBody` hand writes the bodies of `RequestMoreInfo` and `IncorrectInfo` in JSX, and the subjects below are string literals that have to match the ones in the wrapper. They've already drifted: the preview says "reply to the support email included with this message", the template says "reply to this email or reach out to our support team". Two sources of truth for the same words will keep drifting. Either export the subjects from one module both sides import and render the preview from the same React components, or drop the body preview and preview just the subject plus the custom message.

**This file was split in #458.** The entry point is 270 lines of query wiring now, and your +277 would put it past 500, against the under-300 guidance in AGENTS.md section 15. Post rebase this wants to be a new `pending-users-tab/send-email-dialog.tsx` (kebab case, matching that directory per the "match the directory you are already in" rule), with the mutation alongside approve and reject in `use-pending-users-actions.ts`, the trigger in `pending-user-row.tsx` and the dialog rendered from `pending-users-dialogs.tsx`.

PR #357, opened 2026-06-29
feat(rag): agentic retrieval with hybrid search + page citations (#271)

### LINKED ISSUE #271: RAG can miss specific PDF details and answer with overconfident denial
## Summary
A reported chat against an uploaded academic PDF produced an overconfident false negative for a specific detail in the source document. The model initially answered that the article did **not** mention a particular event. When the user later supplied the page number and quoted the relevant sentence, the chatbot corrected itself.

This appears to be a RAG accuracy bug rather than a missing-file/indexing failure: local investigation with a comparable PDF path confirms the relevant text is extractable and chunked, and the exported answer listed all available source chunks for that document. The issue is more likely retrieval/context ordering + chunk adjacency + prompt behavior around negative claims.

## User-visible failure pattern
1. User asks a close-reading question about whether a specific event/detail appears in an uploaded PDF.
2. Assistant answers incorrectly with a confident denial, e.g. "the article does not say...".
3. User supplies the page number and quotes the exact sentence.
4. Assistant then finds/cites the relevant passage and corrects itself.

Expected behavior: the first answer should either find the relevant sentence and answer correctly, or respond with uncertainty instead of a confident denial.

## Local investigation notes
Using the same file-processing path as the app, local extraction and chunking show the likely failure mode:

- PDF text extraction succeeds.
- The relevant passage is present in extracted text.
- Current chunking creates a small number of chunks for the document.
- The reported assistant answer listed all source chunks, suggesting the model likely received the document context but still made an incorrect absence claim.

Current splitter settings in `packages/ai/src/rag-service.ts`:

```ts
new RecursiveCharacterTextSplitter({
  chunkSize: 2500,
  chunkOverlap: 250,
  separators: ["\n\n", "\n", ".", " ", ""],
})
```

The important passage is split in a way that weakens local context:

- One chunk contains the broader example/setup and named context.
- The following chunk starts mid-discussion and contains the exact sentence needed to answer the user's question.

Because `buildRAGContext()` orders chunks by vector distance and formats them in that order, adjacent PDF chunks are not necessarily shown to the model in original document order. The exact sentence may be present, but detached from the previous chunk that explains what it refers to and surrounded by reordered context.

Relevant code paths:

- `packages/ai/src/rag-service.ts`: PDF extraction strips page structure; chunking is plain recursive text splitting.
- `apps/web/src/lib/file-processor.ts`: stores chunks with `chunkIndex`, `content`, `embedding`, `tokenCount`, but no page number or section metadata for PDFs.
- `apps/web/src/server/rag-context.ts`: does pure vector top-k over associated file chunks, orders by vector distance, and sends `[Source: file, Part N]` without neighboring chunks or page labels.
- `apps/web/src/server/routers/chat.ts`: appends RAG context to the professor system prompt, but there is no explicit instruction requiring evidence before saying "the article does not say...".

## Likely causes
1. **Chunk adjacency loss**
   The exact answer sentence can live in a follow-on chunk while the named setup/context is in the previous chunk. Retrieval/context formatting does not preserve neighboring chunks as a unit.

2. **Semantic-only retrieval is weak for exact detail questions**
   Exact terms and quoted phrases should be recoverable through lexical search even when embeddings under-rank the passage.

3. **No page-aware PDF metadata**
   PDF extraction stores text as a single blob. The system cannot prioritize or display page-specific evidence, and it cannot reliably retrieve by page when the user asks `look into page N` except by semantic coincidence.

4. **Context ordering by similarity can harm close reading**
   For small PDFs where many/all chunks fit, preserving source order or grouping hits with neighbors may be better than sending the whole document in semantic-rank order.

5. **Prompt permits overconfident absence claims**
   The model should not say "the article does not say X" unless retrieved evidence is strong enough and the model has checked the relevant context. It should use uncertainty when evidence is incomplete.

## Proposed fixes
- Add hybrid retrieval: combine vector search with lexical/BM25 or PostgreSQL full-text/trigram search for exact terms and quoted phrases.
- Add neighbor expansion: when a chunk is selected, include adjacent chunks from the same file (`chunkIndex - 1`, `chunkIndex + 1`) and preserve local document order inside each group.
- Add page metadata for PDFs if feasible: store `pageNumber` in `fileChunks.metadata` during PDF extraction/chunking or switch to page-aware extraction before chunking.
- For small documents where all retrieved chunks fit, consider sending chunks in `chunkIndex` order after retrieval selection/deduping, not pure similarity order.
- Add a RAG prompt rule: avoid confident negative claims like "the article does not say" unless the answer is directly grounded in retrieved text; otherwise say the retrieved context does not provide enough evidence.
- Add regression coverage with a synthetic fixture that reproduces this pattern: setup in one chunk, exact answer sentence in the neighboring chunk, initial user query asks a close-reading detail without quoting the exact sentence.

## Acceptance criteria
- A close-reading question about a specific event/detail retrieves the exact passage or its neighboring context.
- The answer does not falsely deny the passage when the relevant PDF is attached.
- Source metadata is useful enough to identify the relevant file and preferably page/part.
- Tests cover chunk-neighbor inclusion and/or hybrid retrieval for exact phrase/detail queries.

## Severity
High. This is a user-facing accuracy problem in the core promise of course/file-grounded answers. It is especially damaging because the model gave a confident denial instead of uncertainty.

akhileshrangani4:
## Summary

Fixes #271 — RAG missing specific PDF details and answering with overconfident false denials.

Replaces single-shot static top-k injection with **agentic retrieval**: tool-capable models call retrieval tools mid-turn (hybrid search + page/neighbor lookups), ground answers in what they find, and cite page numbers. Static RAG injection is kept as a fallback. Backed by **page-aware PDF ingestion** and **triple-fusion hybrid retrieval** (vector + Postgres full-text + trigram, fused via Reciprocal Rank Fusion).

## What changed

**Data layer**
- `pg_trgm` extension + FTS (`tsvector`) and trigram GIN indexes on `file_chunks` (`CREATE INDEX CONCURRENTLY`, non-destructive) in `setup-extensions.sql`
- `processingVersion` on `userFiles.metadata`

**Ingestion**
- Page-aware PDF chunking: split on form-feed `\f`, chunk within page boundaries, tag `pageNumber` (fixed `sanitizeText` stripping `\f`)
- Chunk size 2500/250 → 1000/150 for citation precision
- Persist `pageNumber` + stamp `processingVersion`

**Retrieval**
- `reciprocalRankFusion` helper (k=60, weighted)
- Chatbot-scoped triple-fusion hybrid search (vector + tsvector + trigram), quoted-phrase FTS boost

**Agentic layer**
- `supportsTools` flag on `MODEL_REGISTRY` (Gemma marked false — broken tool-calling via OpenRouter); model picker filtered to tool-capable models (legacy saved model still shown)
- 5 AI SDK v6 tools: `search_documents`, `get_page`, `get_context_around`, `list_documents`, `done`
- `streamText` multi-step loop: `stopWhen [stepCountIs(6), hasToolCall("done")]`, `prepareStep` forces search on step 0
- Grounding/anti-denial prompt; sources carry page numbers
- Static RAG path is the fallback (non-tool models, zero-tool-call turns, empty agentic output)

**Safety / UX**
- Authz: `search_documents` scopes model-supplied `fileId` to the chatbot (defense in depth in tool + `hybridSearch`)
- Never-empty answers: the model phrases its own "couldn't find it" via the static fallback (no hardcoded string); truly-empty turns stop the loader cleanly and aren't persisted
- Live agentic status line ("Searching documents…", "Reading page 14…") + page-number source badges (dedup by file+page)
- Lazy reprocess: stale (pre-v2) files re-enqueued non-blocking on chat access

## Test plan
- [x] `npm run check-types`, `npm run lint`, `npm run test` (355 web + 49 ai + 10 db pass; includes new RRF, page-chunking, models, message-sources, retrieval-tools tests)
- [ ] Run migration: `npm run db:setup-extensions` against the DB
- [ ] Live E2E: tool-capable model + multi-page PDF → the previously-failing close-reading question cites a page instead of denying
- [ ] Verify an empty-search question returns a real model sentence (not blank, not canned)

## Notes / follow-ups
- Open-weight tool reliability mitigated via `supportsTools` gating + forced step-0 search + static fallback + `experimental_repairToolCall` + tool-name lowercase normalization
- Known minor: lazy-reprocess can double-publish in the window between enqueue and job start (work is idempotent); no automated test yet for the empty-turn UI path
- `apps/web` and `packages/ai` resolve different `ai` minor versions — one localized boundary cast in `chat.ts` (commented); worth aligning later

### DIFF
--- a/apps/web/src/__tests__/lib/message-sources.test.ts
+++ b/apps/web/src/__tests__/lib/message-sources.test.ts
@@ -1,0 +1,61 @@
+import { describe, it, expect } from "@jest/globals";
+import { dedupeSourcesByFileName, describeSource } from "@/lib/message-sources";
+
+describe("dedupeSourcesByFileName", () => {
+  it("keeps the highest-similarity chunk per file", () => {
+    const result = dedupeSourcesByFileName([
+      { fileName: "a.pdf", chunkIndex: 0, similarity: 0.4 },
+      { fileName: "a.pdf", chunkIndex: 1, similarity: 0.9 },
+    ]);
+    expect(result).toHaveLength(1);
+    expect(result[0]?.similarity).toBe(0.9);
+  });
+
+  it("keeps two pages of the same file as separate badges", () => {
+    const result = dedupeSourcesByFileName([
+      { fileName: "a.pdf", chunkIndex: 0, similarity: 0.8, pageNumber: 3 },
+      { fileName: "a.pdf", chunkIndex: 1, similarity: 0.7, pageNumber: 14 },
+    ]);
+    expect(result).toHaveLength(2);
+    expect(
+      result.map((s) => s.pageNumber).sort((a, b) => (a ?? 0) - (b ?? 0)),
+    ).toEqual([3, 14]);
+  });
+
+  it("collapses same file + same page, keeping highest similarity", () => {
+    const result = dedupeSourcesByFileName([
+      { fileName: "a.pdf", chunkIndex: 0, similarity: 0.5, pageNumber: 3 },
+      { fileName: "a.pdf", chunkIndex: 1, similarity: 0.95, pageNumber: 3 },
+    ]);
+    expect(result).toHaveLength(1);
+    expect(result[0]?.similarity).toBe(0.95);
+  });
+
+  it("treats missing pageNumber as its own key, distinct from a numbered page", () => {
+    const result = dedupeSourcesByFileName([
+      { fileName: "a.pdf", chunkIndex: 0, similarity: 0.5 },
+      { fileName: "a.pdf", chunkIndex: 1, similarity: 0.6, pageNumber: 2 },
+    ]);
+    expect(result).toHaveLength(2);
+  });
+
+  it("returns the input untouched when empty", () => {
+    expect(dedupeSourcesByFileName([])).toEqual([]);
+  });
+});
+
+describe("describeSource", () => {
+  it("strips the Web: prefix and flags web sources", () => {
+    expect(describeSource({ fileName: "Web: example.edu" })).toEqual({
+      isWeb: true,
+      label: "example.edu",
+    });
+  });
+
+  it("returns file sources as-is", () => {
+    expect(describeSource({ fileName: "syllabus.pdf" })).toEqual({
+      isWeb: false,
+      label: "syllabus.pdf",
+    });
+  });
+});
--- a/apps/web/src/__tests__/server/chat-helpers.test.ts
+++ b/apps/web/src/__tests__/server/chat-helpers.test.ts
@@ -1,0 +1,51 @@
+import { describe, it, expect } from "@jest/globals";
+import { clampMaxTokens, describeToolActivity } from "@/server/chat-helpers";
+
+describe("clampMaxTokens", () => {
+  it("returns the default for null/undefined/NaN", () => {
+    expect(clampMaxTokens(null)).toBe(2000);
+    expect(clampMaxTokens(undefined)).toBe(2000);
+    expect(clampMaxTokens(NaN)).toBe(2000);
+  });
+  it("clamps below the minimum up to 100", () => {
+    expect(clampMaxTokens(0)).toBe(100);
+    expect(clampMaxTokens(50)).toBe(100);
+  });
+  it("clamps above the maximum down to 4000", () => {
+    expect(clampMaxTokens(10000)).toBe(4000);
+  });
+  it("passes through values in range", () => {
+    expect(clampMaxTokens(1500)).toBe(1500);
+  });
+});
+
+describe("describeToolActivity", () => {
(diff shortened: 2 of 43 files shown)


akhileshrangani4:
Follow-up (not blocking this PR): the DB/streaming code added here (`hybrid-search` SQL, `chat.ts` streaming loop, `reprocess`, `analytics.deleteConversations`, `file-processor`) shows 0% patch coverage because it's integration-level — unit tests there would be all-mocks. Real coverage needs an integration harness with a throwaway Postgres (pgvector + pg_trgm). Pure logic is covered (`rrf`, `message-sources`, `models`, `retrieval-tool-schemas`, `chat-helpers`, `hasQuotedPhrase` all ~100%). Also deferred: concurrency cap on lazy reprocess to avoid a publish burst when many stale files exist.

PR #378, opened 2026-07-07
feat(auth): auto-sync approved users to Resend audience

### LINKED ISSUE #362: Auto-sync approved users to Resend audience on approval
## Summary

Automatically sync approved Teach Anything users into a Resend audience. Today the admin (Prof Joubin) exports the roster CSV manually and uploads it to Resend, which creates duplicates and is error-prone. We want approval on Teach Anything to push the new user to Resend automatically, with no duplicates.

## Why

- Manual export → manual CSV upload is tedious and repeated on every new signup.
- Re-uploading the full roster duplicates existing contacts; the admin only wants new approvals added.

## Proposed approach

Hook into the existing approval flow. `approveUser()` already runs when an admin approves a user and fires the welcome email:

- `apps/web/src/lib/auth.ts:321` — `approveUser(userId)`

Add a Resend Contacts API call right next to `sendApprovalEmail`, using the same try/catch pattern so a Resend failure never blocks approval:

```ts
// in approveUser(), after sendApprovalEmail
try {
  await resend.contacts.create({
    audienceId: env.RESEND_AUDIENCE_ID,
    email: user.email,
    firstName: user.name ?? undefined,
    unsubscribed: false,
  });
} catch (error) {
  logError(error, "Failed to add contact to Resend audience", { userId: user.id });
}
```

Since we push one contact at approval time, duplicates can't happen. Resend's create-contact is also idempotent per audience+email, so re-runs won't double-add.

## Tasks

- [ ] Create a Resend Audience and add `RESEND_AUDIENCE_ID` to env (validate in `lib/env.ts`).
- [ ] Reuse / instantiate the Resend client (key already wired via `RESEND_FROM_EMAIL` in `apps/web/src/lib/email.ts`).
- [ ] Add `contacts.create` call in `approveUser()` (non-blocking, logged on failure).
- [ ] One-time backfill script: iterate existing `user` rows where `status = 'approved'` and push each into the audience.
- [ ] Tests for the sync helper (mock Resend at the boundary).

## Out of scope (follow-up)

This issue is one-directional: approve → add to Resend. It does NOT:

- Remove a contact from Resend when a user is later rejected/disabled (`rejectUser`).
- Sync unsubscribes from Resend back to Teach Anything.

If we want those, track separately.

## Effort

Small. ~15 lines in `approveUser()` + an env var + a backfill script.

---
Reported by Prof Joubin via email (2026-06-27).

gum1x:
## Summary

Automatically adds users to a Resend audience when an admin approves them, replacing the manual roster-CSV-export → Resend-upload workflow that was tedious and created duplicate contacts.

- New `syncUserToResendAudience` helper called from `approveUser()` — never throws, bounded by a 10s timeout, failures logged
- New optional `RESEND_AUDIENCE_ID` env var (validated in `lib/env.ts`)
- One-time backfill script for already-approved users

Closes #362

## Changes

Each change below states what was done and why that choice was made.

### `apps/web/src/lib/env.ts` + `.env.example` — `RESEND_AUDIENCE_ID`, optional

**Why optional:** it matches the other Resend vars (`RESEND_API_KEY` etc. are `.optional()`), and the whole email stack already degrades gracefully when unconfigured. Making it required would break every existing deploy and local dev setup that doesn't use audiences. The trade-off (a misconfigured prod could silently skip syncing) is covered by a warning log in the helper — see below.

### `apps/web/src/lib/resend-audience.ts` — the sync helper

- **Why a separate module instead of inlining in `auth.ts`:** the issue asks for "tests for the sync helper (mock Resend at the boundary)" — a standalone module is the unit that can be tested with the SDK mocked, and it keeps `auth.ts` focused on auth.
- **Why it never throws (returns `boolean` and logs failures):** the issue requires the Resend call to be "non-blocking, logged on failure" — a Resend outage must never make an admin's approval fail. The boolean gives callers and tests an explicit outcome without exceptions.
- **Why it instantiates its own `Resend` client:** `lib/email.ts` has no client to reuse — email sends are routed through QStash and the client lives in the job route (`app/api/jobs/send-email/route.ts`), which this helper mirrors.
- **Why config is checked at call time (not module load):** mirrors the `checkRateLimit` pattern and lets tests flip configuration between cases; `new Resend()` is a trivial construction with no connection cost.
- **Why the 10s timeout (`Promise.race`):** the Resend SDK sets no fetch timeout, and the tRPC route allows `maxDuration = 300` — without a bound, a hung Resend API would stall the admin's approve action for up to 5 minutes and end in a misleading "Failed to approve user" toast for a user who *is* approved. `Promise.race` is used because the SDK accepts no `AbortSignal`.
- **Why a warning when skipped:** if `RESEND_API_KEY` is set but `RESEND_AUDIENCE_ID` isn't, emails keep working while audience sync silently no-ops — without a log line there would be zero signal that the feature is disabled in prod.
- **Why `unsubscribed` is omitted from the payload:** Resend does not document what a duplicate create does to an existing contact, and its CSV-import path *is* documented as overwriting. Omitting the field guarantees that re-approving a user (or re-running the backfill) can never flip an existing opt-out back to subscribed.
- **Why the full name goes in `firstName`:** Resend contacts have no plain `name` field, and the issue's proposed approach specifies this mapping.

### `apps/web/src/lib/auth.ts` — call site in `approveUser()`

- **Why `approveUser()`:** it is the single choke point every approval flows through (the admin router delegates to it), and it's where the issue's proposed snippet hooks in. The call sits after the approval email so the user-facing effect keeps priority.
- **Why `await` instead of fire-and-forget:** on serverless, the function can be frozen as soon as the response returns, which would drop un-awaited work and silently skip syncs. Awaiting is safe now that the helper is bounded at 10s and never throws.

### `packages/db/scripts/backfill-resend-audience.ts` — one-time backfill

- **Why `packages/db/scripts/`:** it's the established home for one-off operational scripts (`migrate-model-ids.ts`, `seed.ts`, …); this script copies their self-contained dotenv/postgres bootstrap so it runs with plain `npx tsx`.
- **Why raw `fetch` instead of the SDK or the app helper:** `packages/db` has no `resend` dependency and adding one for a one-off script isn't justified (AGENTS.md: don't add dependencies); importing from `apps/web` would invert the package dependency direction. The REST call is 10 lines.
- **Why the 250ms throttle:** Resend's documented default limit is 10 req/s per team; ~4 req/s leaves headroom for concurrent app traffic.
- **Why a 15s `AbortSignal.timeout` + per-user try/catch:** the loop is serial, so without a bound one hung request stalls every user behind it; with per-user containment a single failure is counted and reported instead of aborting the run.
- **Why `process.exitCode` instead of `process.exit()`:** `process.exit()` skips the `finally` block, so the DB connection would never close; `exitCode` lets cleanup run and still reports failure.
- **Why re-running is safe:** the script only creates contacts (never deletes) and omits `unsubscribed`, so the worst case of a re-run is Resend reporting already-existing contacts as failures — no data is lost or overwritten.

### Tests — `apps/web/src/__tests__/lib/resend-audience.test.ts`

- **Why `jest.unstable_mockModule`:** in this repo's ESM setup (ts-jest `useESM` + `--experimental-vm-modules`), `jest.mock()` does not intercept ESM imports; `unstable_mockModule` + dynamic import is the working pattern (same as `transcribe-route.test.ts`).
- **Why mocks live in test-scope variables with implementations set in `beforeEach`:** `jest.config.ts` sets `resetMocks: true`, which strips implementations defined in module factories before each test.

## Library Choices (if applicable)

No new dependencies. Reuses `resend@^4` already in `apps/web`; the backfill script uses Node's global `fetch`.

## Screenshots (if visual changes)

N/A — no UI changes.

## Test Plan

- [x] Unit tests for the helper (`resend-audience.test.ts`, 7 cases): warns + skips when API key or audience ID unconfigured, sends correct payload, omits `firstName` for nameless users, returns `false` + logs on Resend error response, on thrown error, and on timeout (fake timers). Resend SDK mocked at the boundary via `jest.unstable_mockModule`.
- [x] Backfill script fail-fast verified manually: exits 1 with a clear message when `RESEND_AUDIENCE_ID` is missing, before touching the DB or network.
- [x] `npm run lint`, `npm run check-types`, `npm run test` all pass (enforced per-commit by the pre-commit hook).
- Post-deploy: create the Audience in the Resend dashboard, set `RESEND_AUDIENCE_ID`, run `npx tsx packages/db/scripts/backfill-resend-audience.ts` once, then approve a test user and confirm the contact appears in the audience. Recommended first: create one contact twice manually to observe Resend's (undocumented) duplicate-create response.

## Checklist

- [x] PR title follows Conventional Commit format
- [x] Linked to an issue (or context provided)
- [x] `npm run lint` passes
- [x] `npm run check-types` passes
- [x] `npm run test` passes
- [x] No `console.log` statements (script `console.log` follows `packages/db/scripts/` convention)
- [ ] Zod validation on all new tRPC inputs — N/A, no new tRPC endpoints
- [ ] Ownership checks on protected resources — N/A, no new endpoints
- [ ] Screenshots attached for visual changes — N/A
- [x] New dependencies justified in PR description (none added)
- [x] Tests added for new utility/validation logic
- [ ] Database migrations generated — N/A, no schema changes
- [x] I understand and can stand behind every change in this PR


akhileshrangani4:
hey, thanks for this, really solid PR. pushed a bunch of changes straight to your branch so it's ready to go:

- went through the review notes: the timeout actually aborts the request now (it was leaking the in-flight call) and dropped it to 5s since someone's waiting on the approve click, the "api key set but no segment id" case logs an error instead of a silent warn, split the name into first/last, dropped the non-null asserts, and the backfill counts "already present" separately so a clean re-run exits 0
- bundled one small unrelated thing in here so it's not its own PR: a Docs link in the dashboard header next to Support, since we just shipped docs
- bigger one, heads up: Resend deprecated the whole Audiences API in favor of Segments while this was open, so i migrated it over. bumped resend 4 to 6 and used the sdk's contacts api (creates a global contact + segments: [{ id }] now instead of POST /audiences/:id/contacts), and renamed the env var to RESEND_SEGMENT_ID. the sdk was also serializing segments as [{ id }] on the wire, so good thing we're not hand-rolling that
- one gotcha from the resend bump: email.ts imports @react-email/render directly but was only getting it transitively through resend@4, which v6 drops. added it as a proper dep so the build doesn't break

lmk if anything looks off, thanks again for putting this together

PR #380, opened 2026-07-11
feat(chat): migrate chat to AI SDK transport with quiz study tool (Phase 1)

### LINKED ISSUE #377: feat(chat): migrate chat to AI SDK transport + study modes (Phase 1: quiz)
## Summary

Migrate the chat send-path from tRPC subscriptions to the native AI SDK chat transport (`/api/chat` + `/api/chat/shared` Route Handlers, `@ai-sdk/react` `useChat` on the client), and prove it end-to-end with the first render-only study tool, `showQuiz`. Successor to the direction change requested in the #313 review: study modes become native `ai@6` tools instead of a hand-rolled mode registry + custom stream protocol.

Detailed task-by-task plan by @gum1x (`2026-06-18-study-modes-ai-tools.md`, shared by email); it will be revised per the review notes below before execution.

## Phase 1 scope

- Shared `streamChat()` orchestrator replacing the `processMessage` generator (conversation get/create, RAG, token budgeting, history, persistence, analytics), ending in `streamText({ tools }).toUIMessageStreamResponse(...)`
- Two Route Handlers: `/api/chat` (authenticated), `/api/chat/shared` (public + embed), `maxDuration = 300`
- `showQuiz` render-only tool (no `execute`): `inputSchema` is the quiz Zod schema, client renders `<QuizMessage>` from the `tool-showQuiz` message part
- Client: `useChat`/`useChatbot` wrap `@ai-sdk/react` `useChat`; `ChatMessage` renders `message.parts`; `useChatState` deleted
- Persistence: assistant `UIMessage.parts` stored in existing `messages.metadata` JSONB (**no DB migration** — type-widening only, verified via `npm run db:generate` producing nothing)
- tRPC keeps read paths (`getHistory`, `deleteConversation`); the two streaming subscriptions are removed
- `sendReasoning: false` always (reasoning must never leak on public bots)

## Required revisions before execution (plan predates the current pipeline)

The plan's Task 5 ports `processMessage` as of June 18. The function has since gained agentic retrieval (#357) and latency/transport fixes (#370). These must be folded in:

- [ ] `streamChat` passes **both** tool sets: retrieval tools from `createRetrievalTools` (`search_documents`, `get_page`, `get_context_around`, `list_documents`, `done`) **and** `studyTools`. Keep the grounding rule + injected hybrid context from current `chat.ts`, and extend the `useTools` gate so study tools are available on chatbots with no files (today tools are off when `fileIds` is empty; `ragFailureNote` should still disable retrieval tools).
- [ ] **Tool-result privacy**: `toUIMessageStreamResponse` streams tool parts by default. Retrieval-tool parts (raw document chunks) must be filtered out of the UI stream, especially for public/embed bots. Study-tool parts are the ones that should stream.
- [ ] Shared route derives the rate-limit IP via `getTrustedClientIp` (`lib/get-client-ip.ts`), not leftmost `x-forwarded-for`.
- [ ] Keep the `maybeEnqueueReprocess(db, chatbot.id)` fire-and-forget (lazy page-aware re-indexing of old files).
- [ ] Align `ai` versions: add `ai` + `@ai-sdk/react` to `apps/web` **and** bump `packages/ai` to the same `ai` version in one commit (today `apps/web` resolves `ai` through hoisting, which is why `chat.ts` needs `Tool` type casts).
- [ ] Preserve the voice-input props added by #307/#373: `shareToken`, `chatbotId`, `voiceInputEnabled` through `ChatInterface` → `ChatInput`, and `useEmbedVoice` in the embed window.
- [ ] Status/thinking UX: derive the "Searching documents…" status line client-side from streaming tool-input parts; keep a thinking indicator so reasoning/tool phases don't look frozen (`sendReasoning` stays `false`).
- [ ] **Professor dashboard**: render `metadata.parts` in the conversation viewer (Student Chats) so professors see quizzes in their interactive form — read-only `<QuizMessage>`. This is the primary stakeholder requirement and must land before the feature is announced.

## Phase 2 (separate follow-up)

Flashcards, Test (+ open-answer grading via the normal chat stream), Mind Map, Matching — port schemas + widgets from #313, add the tools, extend the system addendum, add render branches.

## Acceptance criteria (Phase 1)

On all three entry points (authenticated page, public share page, embed window):
1. Normal questions stream token-by-token with typing indicator
2. "Quiz me on <topic>" renders an interactive quiz widget grounded in the chatbot's documents
3. RAG sources still appear under answers when enabled
4. Quiz assistant rows persist `metadata.parts` + sources/responseTime; analytics rows carry the right `eventType`
5. Stop mid-stream works; rate limiting intact; no raw document chunks visible in the network stream on public bots
6. Full pre-commit gate green (types, lint, tests)

## Context

- #313 (closed PR) — original study modes implementation; widgets and schemas get ported from it
- Origin review: https://github.com/akhileshrangani4/teachanything/pull/313#pullrequestreview (architectural direction)
- Requested by Prof Joubin for the July newsletter announcement

gum1x:
Closes #377

## Summary

Migrates the chat send-path from tRPC subscriptions to the native AI SDK chat transport (`/api/chat` + `/api/chat/shared` Route Handlers, `@ai-sdk/react` `useChat` on the client) and ships the first render-only study tool, `showQuiz`, end-to-end: model tool call → streamed `tool-showQuiz` part → interactive `<QuizMessage>` widget → persisted in `messages.metadata.parts` → read-only reveal in the professor dashboard. No DB migration (`npm run db:generate` produces nothing — the schema change is type-widening only).

Every item below states **what** was done and **why that choice** — including where the implementation deviates from or goes beyond the plan.

## The #377 revision checklist, item by item

**1. Both tool sets + extended gate** — `streamChat` passes retrieval tools *and* `studyTools`, gated by capability (`stream-chat.ts`):

- **Study tools are always on for tool-capable models, even with zero files.** Why: a system-prompt-only chatbot must still support "quiz me" — under the old gate (all tools off when `fileIds` is empty) quizzes would silently not exist on those bots.
- **Retrieval tools require files AND a healthy RAG pipeline.** Why: a search tool over an empty corpus wastes a round trip and primes the model to answer "I couldn't find that in the materials" when it should just answer. A `ragFailureNote` (embeddings down) takes the same no-retrieval path — tool searches would fail exactly the way the initial retrieval did, so the prompt carries the failure notice instead.
- **The prompt follows the tools**: the grounding rule ("you MUST check the retrieved passages…") and file manifest are only assembled when retrieval tools are actually registered. Why: instructing the model to call tools it doesn't have produces failed tool-call loops.
- **`stopWhen: [stepCountIs(5), hasToolCall("done")]`** — kept the `done` early-exit from #357 (harmless when `done` isn't in the toolset) and bumped the step cap from 4 to 5. Why: a turn can now legitimately spend steps on a retrieval loop *plus* a `showQuiz` call; without the extra step, "quiz me on X" on a document-heavy bot could hit the cap before rendering the quiz.

**2. Tool-result privacy** — a `TransformStream` filter drops retrieval-tool *output* chunks from the UI stream; a persist-time filter strips retrieval parts from `metadata.parts`.

- Output chunks carry only `toolCallId`, not the tool name, so retrieval call IDs are registered from input chunks. **Registration happens on all three input variants (`tool-input-start`/`-available`/`-error`), not just `tool-input-start`.** Why: providers that return tool calls atomically emit `tool-input-available` with no preceding `tool-input-start` — tracking only the latter would let their output chunk slip through, i.e. a raw-document-chunk leak on public bots.
- Tool *inputs* deliberately still stream. Why: they carry no document content (only the model's query/page args) and they power the client status line (item 7) for free.
- Parts are re-filtered before persistence. Why: `metadata.parts` feeds the professor dashboard, which needs text + study parts only; raw chunks should never enter the row even though the dashboard wouldn't render them — defense in depth.
- `sendReasoning: false` on **both** sub-streams (primary and fallback), so reasoning can't leak on public bots through either path.

**3. Trusted client IP** — the shared route keys the per-IP limiter on `getTrustedClientIp` (`x-real-ip`, then rightmost XFF hop), never the leftmost `x-forwarded-for`. Why: the leftmost entry is attacker-controlled; rotating it mints unlimited rate-limit buckets. Beyond the checklist, a **global per-shareToken limiter** was added on top. Why: the per-(IP, token) limiter is still defeatable by a distributed caller rotating real source IPs — the global cap bounds total LLM spend attributable to one shared link regardless of source, set high enough that a classroom of students isn't throttled.

**4. Lazy re-indexing kept** — `maybeEnqueueReprocess(db, chatbot.id)` carries over as the same fire-and-forget (not awaited). Why: it must never add latency to the send-path; failure is tolerable (the reprocess retries on the next message).

**5. `ai` version alignment** — `ai@^6` + `@ai-sdk/react` added to `apps/web` and `packages/ai` bumped to the same `ai` version in a single commit. Why: `apps/web` previously resolved `ai` through hoisting from `packages/ai` (the reason for the old `Tool` type casts); a split-version window between two commits would make every intermediate commit fail type-check, breaking bisect. `npm ls ai` resolves a single `ai@6.0.222` across both workspaces.

**6. Voice input preserved** — `shareToken`/`chatbotId`/`voiceInputEnabled` still flow `page → ChatInterface → ChatInput`, and the embed window keeps `useEmbedVoice` untouched. Why this needed explicit care: `ChatInput`'s voice props have defaults, so dropping one during the prop rewiring would have hidden the mic button with **no type error** — the exact silent failure #377 warned about.

**7. Client-derived status line** — with `sendReasoning: false` and no server status events, the client derives the label from the last in-flight part. Two deliberate choices:

- **Keyed off "the last part is a retrieval part", not part state.** Why: retrieval outputs are filtered server-side (item 2), so retrieval parts never reach `output-available` on the client — keying off state would show "Searching documents…" forever.
- **Reuses the existing `describeToolActivity` helper** rather than a new flat string. Why: the migration had orphaned it (it powered the old server-driven labels), and tool inputs now stream to the client anyway — so the richer labels ("Searching documents for “…”", "Reading page 14…") come back for free instead of leaving dead code behind. Its `get_page` case gained a guard because the client calls it while the tool input is *still streaming*: before `pageNumber` arrives in the partial JSON, the old unguarded `String(args.pageNumber)` would have flashed "Reading page undefined…" (test added).

**8. Professor dashboard** — the Student Chats viewer rehydrates rows through `rowToUIMessage` and renders `<ChatMessage readOnly>`; `QuizMessage` gained a `readOnly` mode showing every question with the correct answer and explanation inline. Why read-only rather than interactive: professors submitting answers into client-side state would go nowhere and pollute the "what did the student see" mental model. Legacy text-only rows fall back to a single text part, so old conversations render exactly as before. Student *answers* stay client-side in Phase 1 (render-only tool, no result round-trip) — persisting them needs an answer-capture path and is deliberately not smuggled into this PR.

## Decisions beyond the checklist (and why)

- **Routes rebuild a trusted user message** (`server/chat/request.ts`): the wire `UIMessage` is validated loosely, then a server-built message forces `role: "user"`, keeps only text parts, and caps length. Why: the client fully controls the transport payload — especially on the unauthenticated `/shared` route, trusting it would allow injecting `role: "system"` turns or planting fake tool parts into history and persistence.
- **`/api/chat` restores the approval-status gate** (non-admin users must be `approved`). Why: the old tRPC path ran through `protectedProcedure`, which enforced this; sessions stay valid after an admin disables an account, so login alone would let a rejected/banned user keep consuming paid inference.
- **Abort + timeout via `AbortSignal.any([req.signal, timeout(300s)])`**: Why: without wiring the request signal, a client Stop or disconnect leaves the model generating server-side (pure wasted spend); the timeout backstops `maxDuration`, which only exists on Vercel.
- **Aborted turns are not persisted** (`onFinish` early-returns; the user message, saved up front, stays). Why: parity, not regression — the old generator's persist code sat *after* the stream loop and never ran on unsubscribe either; persisting half-answers would also pollute the dashboard and analytics.
- **Quiz-only turns persist despite empty `content`**: the persist condition is `content.trim() || hasStudyPart`. Why: a turn whose only output is the quiz has no text, and the old "persist only when text exists" rule would silently drop the quiz from the professor dashboard. Genuinely empty turns are still skipped so reloaded history never shows a blank bubble.
- **Empty-response fallback (#357) kept, with `producedQuiz` counting as a visible answer.** Why: a quiz *is* the answer — falling back to a static no-tools turn after a successful quiz would append a redundant prose answer. A *malformed* quiz also suppresses the fallback deliberately: the no-tools fallback cannot render a quiz, and a prose "quiz" is worse than the inline "Couldn't build the quiz. Try asking again." notice.
- **Malformed / atomic-provider quizzes render a notice, not a blank bubble**: `tool-showQuiz` in `output-error`, and the `dynamic-tool` representation some providers produce, both map to a visible error state. Why: schema-invalid tool inputs are a real model failure mode; a silent empty message reads as a broken app.
- **`metadata.parts` typed `unknown[]` in `packages/db`.** Why: the db package must not import app code or `ai` (layering rule); the app owns the real `UIMessagePart` type and casts on rehydration. Verified type-only: `db:generate` emits no migration.
- **`getHistory`/`deleteConversation` stay on tRPC.** Why: the maintainer review scoped the migration to the streaming send-path; read paths gain nothing from moving. `getHistory` is currently unconsumed but is the intended hook for the future cross-reload rehydration phase (out of scope here — a fresh session per page load matches today's behavior).
- **`maxDuration = 300`** mirrors the prior 5-minute subscription cap rather than inventing a new budget.

## Verification

- Full gate green: `check-types`, `lint --max-warnings 0`, 456 tests across the three packages (pre-commit hook enforced on every commit in the branch).
- New suites: quiz schema, `studyTools` registration, `ui-messages` rehydration (incl. legacy rows and quiz-only empty content), retrieval-tool-name filters, chat-helpers (incl. the partial-input status label).
- SDK contracts verified against the installed `ai@6.0.222` dist rather than docs: the `finish` chunk officially carries `messageMetadata` (so sources/truncated provably reach the client); `convertToModelMessages` converts persisted `output-error` parts into a well-formed call+error-result pair (no provider 400 from history); `abort` is a graceful first-class stream part.
- Manual acceptance criteria from #377 (three entry points, quiz on a zero-file bot, devtools check that no raw chunks appear in the public stream, mic button, stop mid-stream, DB rows) — run per the issue's checklist before merge.

## Known follow-ups (deliberately not in this PR)

- **`request.ts` has no dedicated test suite yet** — it's pure validation logic at the new trust boundary and should get one.
- **The "Cancelled" badge on stopped messages was dropped** in the `ChatMessage` rewrite (the old client marked partial answers with a StopCircle indicator). Server behavior is unchanged; restoring the badge needs a small client-side flag since the aborted fetch never delivers the SDK's `abort` chunk.
- Phase 2 (flashcards, test mode + grading, mind map, matching) repeats the now-proven render-only-tool pattern per mode — separate plan, per #377.


apps/web/src/server/chat/stream-chat.ts
```diff
@@ -0,0 +1,566 @@
...
+      }
+
+      const hasVisibleAnswer =
+        Boolean(primaryText.trim()) ||
+        Boolean(doneAnswer?.trim()) ||
+        producedQuiz;
+
+      let finishReason = await primary.finishReason;
+
+      if (!hasVisibleAnswer && useRetrievalTools && !abortSignal.aborted) {
```
akhileshrangani4: Minor / near-unreachable, just flagging: the empty-response safety net is gated on `useRetrievalTools`. A study-only tool-capable bot (zero files, or RAG unhealthy so retrieval is off) that somehow produces neither text nor a `showQuiz` call gets no fallback: `hasVisibleAnswer` is false, nothing persists, and the client indicator just vanishes with no message. In practice `streamText` basically always emits text when it makes no tool call, and a true no-output turn errors into `executeErrored` (which correctly skips persistence), so this is more of a completeness note than a real gap. Since the study-only path is new here, might be worth a comment noting the net intentionally doesn't cover it.

akhileshrangani4 (commented):
Read the whole diff and ran the gate locally: `check-types` green, 478 web tests pass, retrieval tool names line up with `RETRIEVAL_TOOL_NAMES` (no drift), no dangling refs to the deleted `chat` router / `useChatState`, no stray `console.log`.

This is a genuinely excellent PR. The trust-boundary handling stands out: rebuilding a server-owned user message instead of trusting the wire payload, the approval-status gate restored on `/api/chat`, the trusted-IP + global-per-shareToken rate limiting, and the two-layer retrieval-output privacy (stream filter + persist filter with the atomic `tool-input-available` case handled). The abort/timeout wiring and "don't persist a client-aborted turn" parity are exactly right, and the test coverage on the new logic is thorough. The write-up is one of the best I've reviewed.

Not blocking on code. One thing before merge, and one minor observation inline.

**Before merge:** the branch is currently CONFLICTING with `main` (`mergeStateStatus: DIRTY`). Conflicts are confined to `package.json`, `apps/web/package.json`, and `package-lock.json` from the 1.32.0 release bump on main, so it's mechanical (version + lockfile), not logic. A `git sync` (merge main) + `npm install` to regenerate the lockfile should clear it. Worth re-running the gate after since the `ai@6` resolution touches the lockfile.

The #377 manual acceptance checklist (three entry points, quiz on a zero-file bot, devtools check that no raw chunks appear in the public stream, mic button, stop mid-stream, DB rows) still needs a run before merge per the issue.

apps/web/src/lib/rate-limit.ts
```diff
@@ -171,8 +182,18 @@ export async function checkRateLimit(
...
+  let result: Awaited<ReturnType<Ratelimit["limit"]>>;
+  try {
+    result = await ratelimiter.limit(identifier);
+  } catch (err) {
+    logWarn("Rate limiter check failed; allowing request", {
+      ...context,
+      identifier,
+      error: err instanceof Error ? err.message : String(err),
+    });
+    return { success: true, limit: 0, remaining: 0, reset: 0 };
```
akhileshrangani4: This added fail-open return is the problem. `requireRateLimit` (below) delegates straight to `checkRateLimit`, so this swallow now overrides its fail-closed contract: it only denies when the limiter is `null` (Redis un-configured), not when a configured Redis throws at runtime (partition/timeout), which is exactly the scenario `requireRateLimit`'s comment promises to fail closed on.

Blast radius beyond this PR: `auth.ts:322` (password-change / account-deletion brute-force, commented "fail closed if Redis is unavailable"), `transcribe/route.ts:159/168`, and this PR's own `/shared` global spend cap in `shared/route.ts:41-44`. Pre-PR a thrown `.limit()` propagated and the caller denied; now it silently permits.

Fix: don't let `requireRateLimit` inherit the swallow. Give `checkRateLimit` a `failOpen` flag (default true) and pass `false` from `requireRateLimit`, or inline the `.limit()` call there with its own catch returning `{success:false, reset: Date.now()+60000}`.

apps/web/src/lib/questions.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const mcQuestionSchema = z
+  .object({
+    question: z.string().min(1),
+    options: z.array(z.string().min(1)).min(2).max(4),
+    correct_answer: z.string().min(1),
+    explanation: z.string().min(1),
+  })
+  .superRefine((q, ctx) => {
+    if (!q.options.includes(q.correct_answer)) {
```
akhileshrangani4: This cross-field constraint is a zod runtime refinement, so it is dropped from the JSON schema the model receives for the `showQuiz` tool (I verified: the model-facing schema for `correct_answer` is just `{type:"string", minLength:1}`). The model therefore has no structural signal that `correct_answer` must equal one of `options`, but the SDK still runs the full zod schema at tool-input validation and rejects a mismatch, producing an `output-error` quiz part.

Combined with `producedQuiz` in stream-chat.ts (see other comment), a mismatch means the student sees only the "couldn't build the quiz" notice.

Stronger fix: make the answer structural, e.g. replace `correct_answer: z.string()` with `correct_index: z.number().int()` (0-based into `options`). Then a valid tool call is guaranteed to point at a real option, the model sees the constraint (an integer index), and the whole class of mismatch failures disappears. QuizMessage would compare by index instead of string equality.

apps/web/src/server/chat/stream-chat.ts
```diff
@@ -0,0 +1,566 @@
...
+      } catch {
+        executeErrored = true;
+        return;
+      }
+      const allToolCalls = primarySteps.flatMap((s) => s.toolCalls ?? []);
+      const doneCall = allToolCalls.find((tc) => tc.toolName === "done");
+      const doneInput = doneCall?.input as { answer?: unknown } | undefined;
+      const doneAnswer =
+        typeof doneInput?.answer === "string" ? doneInput.answer : undefined;
+      const producedQuiz = allToolCalls.some(
```
akhileshrangani4: `producedQuiz` counts a `showQuiz` tool call even when its input failed validation (state ends up `output-error`). That makes `hasVisibleAnswer` true, so the empty-response fallback below (`!hasVisibleAnswer && useRetrievalTools && ...`) does not fire. The result is a turn with no text answer and no static-RAG recovery, just the error notice on the client.

Given the `correct_answer` refinement isn't visible to the model (see questions.ts comment), malformed quizzes are more common than they look, so this path is worth hardening. Minimal fix: only count `showQuiz` calls that did not error, e.g. exclude tool calls whose corresponding part is `output-error`, so the fallback still delivers a prose answer when the quiz was the only (failed) output.

apps/web/src/server/chat/stream-chat.ts
```diff
@@ -0,0 +1,566 @@
...
+  const fallbackSystemPrompt =
+    ragResult.ragFailureNote +
+    chatbot.systemPrompt +
+    ragResult.fileManifest +
+    ragResult.contextText;
+
+  // History rows -> UIMessages -> ModelMessages, then append the new message.
+  const historyUiMessages = trimmedHistory.map(rowToUIMessage);
+  const uiMessages: StudyUIMessage[] = [...historyUiMessages, userMessage];
+  const modelMessages = await convertToModelMessages(uiMessages, {
```
akhileshrangani4, line 326: Narrow but reachable: when a chatbot's model is switched to a non-tool model, `modelCanUseTools` is false and `tools = {}`, but persisted history can still contain completed `tool-showQuiz` parts (state `output-available`). `convertToModelMessages` won't orphan them (`ignoreIncompleteToolCalls` doesn't apply, they're complete) and will emit provider tool-call + tool-result messages into a request that declares no tools, to a model that may not support tool calling. Depending on the provider that is a 400 that breaks every subsequent turn in the session.

Trigger: study bot on a tool-capable model, a quiz turn persists, admin switches the bot to a non-tool model, student sends another message in the same session. Consider sanitizing history when `!modelCanUseTools`: in `rowToUIMessage` (or just before this call) drop or down-convert `tool-*` parts to text so a non-tool model never receives tool-call messages.

akhileshrangani4 (changes requested):
Follow-up after a deeper multi-agent pass (security/trust-boundary, streaming-correctness, client/UI reviewers each reading only their slice with fresh context). The architecture holds up under the second look, and the trust-boundary + privacy design was independently confirmed sound. Two things I want to flag before merge though, including one regression my first pass got wrong.

Correction to my earlier review: I said the `checkRateLimit` try/catch was pre-existing. It is not. `git diff` confirms this PR added it, and it changes security-critical behavior. Details inline.

### Should fix before merge

1. **`requireRateLimit` now fails open on runtime Redis errors** (rate-limit.ts). This PR added a try/catch to `checkRateLimit` that returns `{success:true}` when `ratelimiter.limit()` throws. `requireRateLimit` delegates to it, so its fail-closed contract now only holds when Redis is un-configured, not when a configured Redis errors at runtime (the exact case its doc-comment names). Blast radius is beyond this PR: `auth.ts:322` (password-change / account-deletion brute-force, commented "fail closed if Redis is unavailable"), public transcription, and this PR's own `/shared` global spend cap. Inline with a fix.

2. **Quiz `correct_answer` constraint never reaches the model, and a malformed quiz suppresses the fallback** (questions.ts + stream-chat.ts). The `.superRefine` that enforces `correct_answer in options` is a zod runtime refinement, so it is stripped from the JSON schema the model sees (verified: the model-facing schema for `correct_answer` is just `{type:string, minLength:1}`), yet the SDK still rejects a mismatch at validation time. So the model can routinely emit a quiz that fails validation, lands as `output-error`, and because `producedQuiz` counts an errored `showQuiz` call as a visible answer, the empty-response fallback is suppressed. Net: the student gets only the amber "couldn't build the quiz" notice with no answer and no retry. Inline with two fix options.

### Worth a look (narrower)

3. **Persisted `showQuiz` history sent to a non-tool model** (stream-chat.ts). Narrow trigger (admin switches a study bot to a non-tool model mid-session) but can 400 the provider on the next turn. Inline.

### Minor / follow-ups (non-blocking)

- **A11y:** the read-only quiz on the professor dashboard marks the correct option via `aria-label` on a `<div>` (QuizMessage.tsx readOnly branch) plus green color + an `aria-hidden` check. Non-interactive `<div>` `aria-label` is not reliably announced, so a screen-reader user cannot tell which option is correct. Use an `sr-only` text node instead. Trivial and worth doing.
- A `showQuiz` part stuck in `input-streaming` at persist time (timeout mid-input) writes an invisible, unrenderable ghost row (still counted by `hasStudyPart`). Exclude non-input/output-available study parts from `hasStudyPart`.
- Study-only tool-capable bot (zero files) has no empty-response safety net; consider gating the fallback on `modelCanUseTools` rather than `useRetrievalTools` (I flagged this in the first review too).
- Quiz-only assistant turns export as a blank `[N] Assistant:` line (export-chat.ts `extractText` sees no text parts). Consider a `[Interactive quiz: <title>]` placeholder.
- Quiz-only rows count as ~0 tokens in the history budget (empty `content`) while carrying real tool-call JSON; low impact given the 50-row cap.
- `deriveStatusLine`'s `if (!isStreaming)` branch is dead (only rendered while streaming); `done` is in `RETRIEVAL_PART_TYPES` but has no `describeToolActivity` case (cosmetic).

Gate is still green locally (check-types + 478 web tests). Merge conflict with main (package.json / lockfile from the 1.32.0 bump) still needs a `git sync` + `npm install`.

akhileshrangani4:
hey, tested this locally and ended up pushing a bunch of commits straight to the branch instead of doing another review round, faster that way. summary of what changed:

**fixes**
- mic button spun forever on localhost. `useVoiceRecorder` never reset `unmountedRef` on mount, so StrictMode's dev double-mount left it stuck true and the post-getUserMedia guard bailed before recording ever started. one line fix, prod was never affected
- quiz had no loading state while the model builds it, added a "Building your quiz..." skeleton for the input-streaming phase

**student answers (the big one)**
- answers were render-only, nothing persisted. added a `study_tool_responses` table + `POST /api/study-response` (and `/shared`) that resolves the conversation from the session, checks the toolCallId was actually shown in it, and grades server side (client score is never trusted). the quiz auto-submits when the student hits finish, retakes get their own attempt
- professor dashboard now shows what students actually picked per attempt, paginated one attempt at a time, opens on the latest. students get the same viewer on their score screen so they can page back through retakes
- chat export (text + json) includes the questions, correct answers, and the student's attempts + scores
- the model now gets a small results note each turn ("attempt 1 scored 2/5", or "shown but not answered yet") so it can actually react to how the student did

**kept it generic for phase 2**
- the table is just toolName + jsonb response, and the endpoint dispatches to a per-tool handler registry. adding flashcards/test/mindmap later = register a handler + build the component. no schema or endpoint changes needed

**hardening (ran a review pass over my own changes too)**
- quiz_title gets sanitized + length-capped before it lands in the system prompt via the results note. a student could steer the model into a title with embedded newline instructions, which is a real injection path into the system channel
- a stopped stream now still persists the turn if it carries a completed quiz, otherwise the rendered quiz stays interactive but every submission for it 404s forever
- attempt POSTs retry transient 404/5xx with backoff, since a fast student can finish a short quiz before onFinish commits the assistant row
- unique index on (conversation, toolCallId, attempt) + conflict retry so attempt numbers are race-safe

two migrations (0017 table, 0018 unique index), both purely additive, CI applies them via db:migrate on release. full gate is green (types, lint, 526 tests, prod build) and i tested the whole flow locally across all three entry points. good work on the base migration btw, the trust-boundary stuff held up really well under review

akhileshrangani4 (approved):

PR #448, opened 2026-08-21
fix: Bug Fixes in file uploads and quiz

gum1x:
- PDFs could not be uploaded at all:  the PDF library crashed before reading the file
- Uploads got stuck at 40% and then failed, and a killed job locked the file permanently
- The quiz bot wrote quizzes out in plain text (answers included), generated a second unrelated answer, and ignored the chapter a student named
- Failure messages were too broad to act on

# Changes
## 1) No one could upload PDFs

The library that reads PDFs has a bug: it contains leftover debugging code that tries to open a sample file the moment it loads. Where it's looking for the sample file doesn't exist, so the library crashes before it actually touches the file.

To fix this, we make it load that library through a side door, so it skips the broken debugging code.

## 2) Files uploading get stuck at 40%, then fail

The upload pipeline had no time limit set, so the uploads were getting cut off. To fix this, I made it have the same five-minute limit as all the other jobs.n addition to this, a job that got killed locked the file. When processing starts, the file gets marked as in progress, so two jobs can't grab it at once. But if one of the jobs dies, it doesn't get unmarked  so the file stays locked forever and every retry gives up on it. There is now a sweep that clears an abandoned lock so the file can be retried.

## 3) Quiz bugs

Llama 4 Maverick doesn't use the proper channel for building quizzes. This was already known, but the recovery step wasn't properly built  it gave up when the model used six questions instead of five, or a differently named answer field, and dumped the raw text on screen (which contains the answers). The system now uses the same repair path as the normal route: it trims extra questions and fixes renamed answer fields.

The history didn't match what the students saw, because the code basically checked "did the model say anything?" by looking at only the last step of its work. If the bot answered early and then did one more step, it looked like it had said nothing, so the app generated a second, different answer and stuck it on the end. That's where the "What topic are you interested in?" reply came from. Also, if a student hit Stop on a response that was still working, the whole reply was thrown away and never saved. To fix this, I made it check all the steps instead of just the last one.

The chapter instructions were also being ignored, because the instructions sent to the bot ended with "base questions on the course material above", pointing at whatever got pulled up automatically, with nothing telling it to respect the chapter the student named. I added that.

In addition to this, the failure messages were overly broad, so I made them less broad. A scanned PDF, a corrupt file and a real outage all reported the same generic "internal error", which is a large part of why this went undiagnosed for so long.

## Test Plan

- [x] Full pipeline run against a real Postgres + pgvector database with a 16-file corpus: real PDFs from five different producers, plus truncated, corrupt, empty, mislabelled, non-English, very long, and legacy-format files — all reach a proper end state with a sensible message
- [x] Verified the PDF fix by A/B: the same file fails to process before the change and completes after it
- [x] Edge cases: file deleted from storage mid-job, job killed mid-embedding, retrying a file uploaded weeks ago, student pressing Stop mid-quiz, quiz cut off by the token limit, model emitting six questions or a renamed answer field
- [x] Checked that ordinary answers containing JSON, code blocks, LaTeX or braces still stream through untouched and are not mistaken for a quiz
- [x] Manual check after deploy: upload a text PDF and a Word file, run a chapter-scoped quiz on the Critical Theory bot, and compare against the Shakespeare bot
- [x] Manual check after deploy: press Retry on a file that previously failed

### DIFF
--- a/apps/web/src/__tests__/lib/crawled-page-files.test.ts
+++ b/apps/web/src/__tests__/lib/crawled-page-files.test.ts
@@ -1,0 +1,45 @@
+import { describe, it, expect } from "@jest/globals";
+import { PgDialect } from "drizzle-orm/pg-core";
+import {
+  excludeCrawledPages,
+  isCrawledPagePath,
+} from "@/lib/crawled-page-files";
+
+/**
+ * `userFiles` mixes uploads and crawled pages, and `storagePath` is the only
+ * thing telling them apart. Both forms of the check are pinned here because
+ * they guard different layers -- the SQL one keeps crawled pages out of the
+ * Files tab and out of `sweepStaleFiles`, the predicate one keeps the guarantee
+ * if a future caller hands a row straight to `isStaleFile`.
+ */
+describe("isCrawledPagePath", () => {
+  it("recognizes a crawled page by its URL storagePath", () => {
+    expect(isCrawledPagePath("https://example.edu/syllabus")).toBe(true);
+    expect(isCrawledPagePath("http://example.edu/syllabus")).toBe(true);
+  });
+
+  it("treats an uploaded object key as an upload", () => {
+    expect(isCrawledPagePath("abc123/9f1c-4e2b.pdf")).toBe(false);
+    // Uploaded paths are `{userId}/{fileId}`, so the letters can appear without
+    // the path being a URL.
+    expect(isCrawledPagePath("abc123/http-notes.pdf")).toBe(false);
+  });
+
+  it("does not throw on a missing path", () => {
+    expect(isCrawledPagePath(null)).toBe(false);
+    expect(isCrawledPagePath(undefined)).toBe(false);
+  });
+});
+
+describe("excludeCrawledPages", () => {
+  it("renders as a negated prefix match on storage_path", () => {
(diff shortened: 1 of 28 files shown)


akhileshrangani4 (dismissed):

akhileshrangani4 (commented):
.

akhileshrangani4:
Pushed a couple of extra fixes onto this branch:

- `sweepStaleFiles` was sweeping crawled pages too. They live in `userFiles` as well and a re-crawl sets them to `processing` without a progress stamp, so the sweep dated them from `createdAt` and failed live re-crawls. Moved `excludeCrawledPages` out of `list.ts` into `lib/crawled-page-files.ts` and applied it to the sweep (plus `isStaleFile` itself), and ordered the candidate read oldest-first so the 200 cap isn't an arbitrary slice.
- `sanitizeProcessingError` matched `includes("password")`, which caught pdf.js's "No password given" but missed "Incorrect Password", so wrong-password uploads got the generic internal-error message. Word-bounded and case-insensitive now.
- Moved the `parseQuizFromText(content)` call into the client-abort branch in `stream-chat`, since that's the only place the result is read.

Tests added for all three. Full suite, lint and check-types pass. Rest of the PR looks good, nice work on the pdf-parse root cause.

PR #461, opened 2026-08-25
fix: harden pre-existing job, stream, upload and auth paths

akhileshrangani4:
Stacked on #458. These are **pre-existing** issues found while reviewing that refactor, each verified byte-identical to `main` before the refactor touched it. None is a regression from #458; they are kept out of it so that PR stays a reviewable "behavior preserved" change with its own rollback.

Review after #458 lands, or set the base to `main` once it does.

## What changed

**`fix(crawl)` — the one that could lose data.** `processCrawlPage` claimed a page with `UPDATE ... WHERE id = ?` and no status predicate, so two QStash deliveries of the same page both proceeded, both read `userFileId` as null, and both inserted a `userFiles` row. The loser was orphaned carrying chunks and chatbot associations nothing points at. The claim is now conditional (`pending` or `failed`, so a retry still recovers a transient failure) and bails on zero rows. The claimable set is exported and pinned against the full `crawled_page_status` enum, so adding a status forces a decision rather than defaulting into the refused set.

Refusing `processing` cannot wedge a page: `sweepStaleCrawls` already recovers stuck `pending`/`processing` pages. Same commit stops generating embeddings *inside* the transaction, which held the source share lock and a pooled connection across provider network calls while discovery ran up to 20 page jobs concurrently. The transaction still covers delete + insert + status atomically.

**`fix(chat)`** — `await forward(...)` sat outside `runPrimaryTurn`'s try, so a writer or transform failure escaped without setting terminal error state and persistence could still call the turn successful. Also cancels the reader instead of only releasing its lock, so a failure stops upstream generation rather than burning tokens on undeliverable output.

**`fix(auth)` session codes** — `gateSessionCreation` raised `ACCOUNT_PENDING` / `ACCOUNT_REJECTED` and then swallowed them in its own catch, so every blocked login looked identical to a database outage. Rethrows `APIError`, keeps swallowing genuine failures. **Changes what the login page can display.**

**`fix(upload)`** — the local-upload route was gated only on `isLocalStorageMode()`, an absence-of-config check, so a prod deploy missing Supabase credentials left an authenticated unbounded `arrayBuffer()` write-to-disk endpoint live on `requireApiSession` (pending/rejected accounts included). Now refuses in production, requires an approved user, and caps the body.

**`fix(auth)` domain matching** — the allowlist used `emailDomain.endsWith(entry)` against the domain with its `@` attached. `@gwu.edu` anchored, but the bare `gwu.edu` form the admin UI accepts also matched `attacker@evil-gwu.edu`. Now matches on a label boundary. `.edu` / `edu` stay TLD wildcards. **Tightens who can register.**

## Verification

Typecheck, lint, and the full suite green. 908 → 919 tests; every behavior change here is covered by a test asserting the new behavior, and the two tests that previously pinned the old permissive behavior were updated deliberately rather than deleted.

## Deliberately not here

Four findings need a schema change or real design work and are filed as issues instead: the file-processor run token, the finalize-upload outbox, the user-deletion tombstone, and per-conversation turn ordering. AGENTS.md §15 forbids schema changes without a migration plan.


akhileshrangani4:
Companion issues for the four findings deliberately left out of this PR (schema change or design work, per AGENTS.md §15): #462 file-processor run token + staged chunk swap, #463 finalize-upload outbox, #464 user-deletion tombstone, #465 per-conversation turn ordering.

PR #496, opened 2026-09-07
fix(crawler): keep attach/detach pending until the source list refetch lands

### LINKED ISSUE #481: web-sources: attach/detach stop showing pending through the refetch, so the picker re-enables on stale data
Found while reviewing #458. Small, but it is a genuine behavior change introduced by that PR rather than a pre-existing issue, and the fix needs a decision so it was left out.

## What changed

Before #458, the dashboard page wired attach/detach with an **expression-bodied** arrow:

```ts
onSuccess: () => refreshSources(),   // returns the invalidate() promise
```

react-query awaits whatever `onSuccess` returns before dispatching success. From `@tanstack/query-core/build/modern/mutation.js`:

```js
await this.options.onSuccess?.(...)
// ...then
this.#dispatch({ type: "success" })
```

So `isPending` used to stay true through the list refetch.

The extracted hook fires it without awaiting:

`apps/web/src/hooks/use-crawler-mutations.ts:82-83, 93-94`

```ts
onSuccess: () => {
  void refresh();
  ...
}
```

`isPending` now clears as soon as the server responds, before the refetched list arrives.

## Why it matters

`dashboard-web-source-table/row-cells.tsx` gates the chatbot picker on `disabled={isPending}`. The menu items re-enable while the row is still rendering pre-mutation attachments, so a user can click again against state that is about to change under them.

Minor in impact. Filed because it is silent, and because the hook's own doc comment currently asserts the opposite:

`use-crawler-mutations.ts:42-43`

```
* Extra refresh for attach/detach (...). Not awaited.
```

That is accurate for the `WebSourcesTab` consumer and wrong for the dashboard one.

## Why it is not a one-line fix

`return refresh()` is the obvious change, but the two consumers genuinely differed before the refactor:

- **dashboard page**: expression-bodied, so awaited.
- **WebSourcesTab**: block-bodied, so not awaited.

Making the shared hook await would fix the dashboard and change `WebSourcesTab`. So it needs either a per-mutation opt-in (e.g. `awaitRefreshOn: ["attach", "detach"]`, or an `awaitRefresh?: boolean` on the options object), or a deliberate call that both consumers should now await, with the doc comment corrected either way.

Same pattern is worth a glance on the other five mutations in that hook, though those were block-bodied at both call sites before and after, so they are unchanged.

## Scope

Only `attach` and `detach`. `recrawl`, `cancelCrawlSource`, `toggleCrawlSource`, `removeCrawlSource` are unaffected; `renameCrawlSource` correctly awaits for `WebSourcesTab` and not for the dashboard, matching each old call site.

gum1x:
The attach/detach pending thing you found while reviewing #458. basically the extracted hook stopped holding `isPending` through the list refetch, and i think the fix is small but it needed a call on the two consumers, so i want to walk through what i did and why.

## summary

so before #458 the dashboard page wired attach and detach with an expression body arrow, `onSuccess: () => refreshSources()`, which returns the `invalidate()` promise. react query awaits whatever `onSuccess` returns before it dispatches success, so `isPending` used to stay true until the refetched list came back. the extracted `useCrawlerMutations` hook fires the refresh with `void` instead, so `isPending` clears as soon as the server responds, and the row's chatbot picker enables again while it's still showing the old attachments. one more click there fires against state that's about to change underneath it.

* attach and detach `onSuccess` now return the `refresh()` promise instead of voiding it
* the `refresh` option is tightened from `() => void | Promise<unknown>` to `() => Promise<unknown>`
* new hook test under `__tests__/hooks` that fails on the old code
* closes #481

## changes

so first i wanted to make sure the mechanism was actually real and not just a reading of the code, and i'm pretty sure it is. tanstack query core 5.100.8 awaits `onSuccess` before dispatching success (line 123 of the mutation file), and the trpc react query 11.17.0 `useMutation` wrapper returns your `onSuccess` result straight through, its default `mutationSuccessOverride` is just `options => options.originalFn()` and we don't configure an override in the trpc client. and the gate on the dashboard is `page.tsx` feeding `attach.isPending || detach.isPending` into the `disabled` prop on every checkbox item in `ChatbotAttachButton`, which is the thing that now opens early.

the decision part. the two call sites genuinely differed before #458, the dashboard awaited and `WebSourcesTab` didn't, so a totally faithful restore would need a per consumer flag like `awaitRefresh`. i went with awaiting in both instead. my reasoning is basically that `WebSourcesTab` not awaiting was an accident of block body syntax rather than a design choice, and it has the exact same stale state window, its attach buttons and the per row remove buttons are gated on the same `isPending` values. a flag whose only reason to exist is that one call site used to be an expression body didn't feel worth carrying. so now the hook's contract is one sentence, attach and detach hold `isPending` until the refetched list lands.

in the hook itself:

* attach and detach `onSuccess` still call `refreshAttachments?.()` and the success toast synchronously first, then `return refresh()`. so toast timing in `WebSourcesTab` is unchanged.
* `refresh` is now typed `() => Promise<unknown>`. both existing callers already return promises (`utils.crawler.getAllCrawlSources.invalidate()` and `useQuery().refetch`), so it compiles today. the point is that a future block body callback like `() => { void invalidate(); }`, which is the exact shape that caused this, becomes a compile error instead of a silent revert. it also matches how `refreshAfterRename` was already typed.
* the doc comment on `refresh` now says which mutations await it. the old one asserted the opposite, which is what made this silent.

in the new test:

* it mocks the trpc client and sonner with `jest.unstable_mockModule`, same pattern as the user deletion test, and captures each `useMutation`'s options. the hook has no react hooks of its own besides the mocked trpc ones, so i just call it as a plain function, no renderer needed.
* for both `attachToChatbot` and `detachFromChatbot` it checks that `onSuccess` doesn't settle until `refresh` resolves, and that the toast and `refreshAttachments` fire before the refresh settles. i stashed the hook change and reran it, and the two "holds pending" cases go red on the old code while the toast ordering ones stay green, so it does catch the actual regression.

other than that i left a few things alone on purpose:

* `toggleCrawlSource` and `removeCrawlSource` have the same shape of window, `EnabledSwitch` can snap back to the stale value and enable again before the refetch. but they were block body at both call sites before and after #458, so that's preexisting rather than part of this regression, and i didn't want to widen the pr past what the issue scoped. i think it deserves its own issue.
* `renameCrawlSource` already awaits for `WebSourcesTab` through `refreshAfterRename` and not for the dashboard, matching both old call sites, so untouched.

and two tradeoffs of awaiting that i think are worth knowing about:

* the query client doesn't override `retry`, so a refetch that fails after a successful attach retries 3 times with backoff, something like 7 seconds or so, and the controls stay disabled that whole time. that's identical to what the dashboard did before #458, but it's new for `WebSourcesTab`.
* `WebSourcesTab`'s bulk "remove selected" runs `detach.mutateAsync` one id at a time through `runSequentially`, so each step now also waits for a refetch. removing n sources costs roughly n extra refetch round trips, in exchange for the remove buttons staying disabled for the whole run. if that ever gets noticeable, a refresh free detach for the bulk path with one refetch after the loop would fix it, but that felt like a separate change.

open questions i'd want your read on:

1. do you want toggle and remove to await too so the hook contract is uniform? i can do that in this pr or a follow up, whichever you prefer.
2. toast before or after the await? rename in the same hook toasts after its awaited refresh. i kept attach and detach toasting first because that's what `WebSourcesTab` did before, but i could see the argument for matching rename.
3. is the bulk remove slowdown fine for now, or do you want the refresh free detach in here?
4. i haven't clicked through it on a throttled network yet, so the manual check in the test plan is still open.

## library choices

no new dependencies.
## test plan

* [x] lint passes with zero warnings
* [x] type check passes
* [x] tests pass across the whole monorepo, roughly 940 in web plus the db and ai packages
* [x] new test goes red against the old hook (verified by stashing the hook change)
* [x] verified the whole chain in installed code: hook returns the promise, trpc wrapper returns it, react query awaits it, `page.tsx` passes it, `ChatbotAttachButton` gates on it
* [ ] manual: on the dashboard web sources page, open a row's chatbot picker on a throttled network and attach. the items should stay disabled until the count and checkbox update, with no window where the old state is clickable
* edge cases i thought about: `refresh` rejecting (neither `refetch()` nor `invalidate()` rejects by default, and it's the same semantics the dashboard had before #458), and a sync throw in `refreshAttachments` or the toast (not possible with the current callers)

## checklist

* [x] pr title follows conventional commit format
* [x] linked to an issue (closes #481)
* [x] `npm run lint` passes
* [x] type check passes
* [x] `npm run test` passes
* [x] no `console.log` statements
* [x] zod validation on new trpc inputs, n/a, no new inputs
* [x] ownership checks on protected resources, n/a
* [x] screenshots for visual changes, n/a
* [x] new dependencies justified, none added
* [x] tests added for new logic
* [x] database migrations if schema changed, n/a
* [x] i understand and can stand behind every change in this pr

### DIFF
--- a/apps/web/src/__tests__/hooks/use-crawler-mutations.test.ts
+++ b/apps/web/src/__tests__/hooks/use-crawler-mutations.test.ts
@@ -1,0 +1,87 @@
+import { jest, describe, it, expect } from "@jest/globals";
+
+type MutationOptions = {
+  onSuccess?: (...args: unknown[]) => unknown;
+};
+
+/** `useMutation` options captured per procedure name (attachToChatbot, …). */
+const mockCaptured = new Map<string, MutationOptions>();
+
+jest.unstable_mockModule("@/lib/trpc", () => ({
+  trpc: {
+    crawler: new Proxy(
+      {},
+      {
+        get: (_target, name) => ({
+          useMutation: (options: MutationOptions) => {
+            mockCaptured.set(String(name), options);
+            return {
+              mutate: () => undefined,
+              mutateAsync: async () => undefined,
+              isPending: false,
+              variables: undefined,
+            };
+          },
+        }),
+      },
+    ),
+  },
+}));
+jest.unstable_mockModule("sonner", () => ({
+  toast: { success: jest.fn(), error: jest.fn() },
+}));
+
+const { useCrawlerMutations } = await import("@/hooks/use-crawler-mutations");
+const { toast } = await import("sonner");
+
+function deferred() {
+  let resolve!: () => void;
+  const promise = new Promise<void>((r) => {
+    resolve = r;
+  });
+  return { promise, resolve };
+}
+
+const flushMicrotasks = () => new Promise((r) => setTimeout(r, 0));
+
+describe.each(["attachToChatbot", "detachFromChatbot"])(
+  "useCrawlerMutations %s onSuccess",
+  (procedure) => {
+    it("holds the mutation pending until the list refresh settles", async () => {
+      const refresh = deferred();
+      useCrawlerMutations({ refresh: () => refresh.promise });
+
+      let settled = false;
+      const result = Promise.resolve(
+        mockCaptured.get(procedure)?.onSuccess?.(),
+      ).then(() => {
+        settled = true;
+      });
+
+      await flushMicrotasks();
+      expect(settled).toBe(false);
+
+      refresh.resolve();
+      await result;
+      expect(settled).toBe(true);
+    });
+
+    it("fires the toast and attachment refresh before the list refresh settles", () => {
+      const refresh = deferred();
+      const refreshAttachments = jest.fn();
+      useCrawlerMutations({
+        refresh: () => refresh.promise,
+        refreshAttachments,
+        attachSuccessMessage: "attached",
+        detachSuccessMessage: "detached",
+      });
+
+      void mockCaptured.get(procedure)?.onSuccess?.();
+
+      expect(refreshAttachments).toHaveBeenCalledTimes(1);
+      expect(toast.success).toHaveBeenCalledWith(
+        procedure === "attachToChatbot" ? "attached" : "detached",
+      );
+    });
+  },
+);
--- a/apps/web/src/hooks/use-crawler-mutations.ts
+++ b/apps/web/src/hooks/use-crawler-mutations.ts
@@ -34,10 +34,12 @@ export interface CrawlerMutations {
 
 interface UseCrawlerMutationsOptions {
   /**
-   * Refreshes the source list after every successful mutation. Fired without
-   * awaiting, matching the original inline wiring.
+   * Refreshes the source list after every successful mutation. Attach/detach
+   * await it so `isPending` holds until the refetched list lands (the row's
+   * chatbot picker is gated on it); the other mutations fire it without
+   * awaiting.
    */
-  refresh: () => void | Promise<unknown>;
+  refresh: () => Promise<unknown>;
   /**
    * Extra refresh for attach/detach (e.g. invalidate the chatbot's
    * attachable-sources list). Not awaited.
@@ -80,22 +82,22 @@ export function useCrawlerMutations(
 
   const attach = trpc.crawler.attachToChatbot.useMutation({
     onSuccess: () => {
-      void refresh();
       refreshAttachments?.();
       if (attachSuccessMessage) {
         toast.success(attachSuccessMessage);
       }
+      return refresh();
     },
     onError: (error) => showError("Failed to attach", error),
   });
 
   const detach = trpc.crawler.detachFromChatbot.useMutation({
     onSuccess: () => {
-      void refresh();
       refreshAttachments?.();
       if (detachSuccessMessage) {
         toast.success(detachSuccessMessage);
       }
+      return refresh();
     },
     onError: (error) => showError("Failed to remove", error),
   });


apps/web/src/hooks/use-crawler-mutations.ts
```diff
@@ -34,10 +34,12 @@ export interface CrawlerMutations {
...
+   * Refreshes the source list after every successful mutation. Attach/detach
+   * await it so `isPending` holds until the refetched list lands (the row's
+   * chatbot picker is gated on it); the other mutations fire it without
+   * awaiting.
    */
-  refresh: () => void | Promise<unknown>;
+  refresh: () => Promise<unknown>;
   /**
    * Extra refresh for attach/detach (e.g. invalidate the chatbot's
    * attachable-sources list). Not awaited.
```
akhileshrangani4: This is the one I'd want changed before merge, and it's the surface the fix doesn't quite reach.

In `WebSourcesTab` the attach button list isn't driven by `refresh` at all. `unattached` is built from `attachable` (`WebSourcesTab.tsx:150`), which comes from `trpc.crawler.getAttachableSources.useQuery` (`:59`), which is `refreshAttachments`, still fire and forget. So on that surface `AttachExistingSources.tsx:54`'s `disabled={isAttaching}` can still re-enable over a stale `unattached` list, which is the same bug the PR is fixing, one query over.

In practice the window is tiny because both refetches start in the same tick and ride the same `httpBatchLink` batch, so they land together. That's why I'm not blocking on it. But it's luck rather than design, and the doc comment right here says "Not awaited" while the contract you're establishing is that attach and detach hold pending until the list is right.

```ts
refreshAttachments?: () => Promise<unknown>;
// ...
onSuccess: () => {
  const attachments = refreshAttachments?.();
  if (attachSuccessMessage) toast.success(attachSuccessMessage);
  return Promise.all([refresh(), attachments]);
},
```

`WebSourcesTab` already passes `() => utils.crawler.getAttachableSources.invalidate()`, which returns a promise, so this compiles as is.

apps/web/src/hooks/use-crawler-mutations.ts
```diff
@@ -34,10 +34,12 @@ export interface CrawlerMutations {
...
   /**
-   * Refreshes the source list after every successful mutation. Fired without
-   * awaiting, matching the original inline wiring.
+   * Refreshes the source list after every successful mutation. Attach/detach
+   * await it so `isPending` holds until the refetched list lands (the row's
+   * chatbot picker is gated on it); the other mutations fire it without
+   * awaiting.
    */
-  refresh: () => void | Promise<unknown>;
+  refresh: () => Promise<unknown>;
```
akhileshrangani4, line 47: Worth one sentence in the doc comment saying `refresh` must not reject.

query-core awaits `onSuccess` inside the try, so a rejection here doesn't just fail the refetch, it flips a successful mutation into the error branch and fires `onError`. You'd get a "Failed to attach" toast on an attach that worked, and `mutateAsync` would reject, which `runSequentially` counts as a failed removal.

Not reachable today, I checked both callers: `refetchQueries` swallows per query with `promise.catch(noop)` (`queryClient.js:330`) and `#executeFetch` does the same unless `throwOnError` (`queryObserver.js:245`). But tightening the type to `() => Promise<unknown>` is an invitation for someone to pass a promise that can reject, so the constraint is worth writing down next to the type.

apps/web/src/__tests__/hooks/use-crawler-mutations.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+function deferred() {
+  let resolve!: () => void;
+  const promise = new Promise<void>((r) => {
+    resolve = r;
+  });
+  return { promise, resolve };
+}
+
+const flushMicrotasks = () => new Promise((r) => setTimeout(r, 0));
```
akhileshrangani4: Nit: this is a macrotask hop, not a microtask drain. `setTimeout(r, 0)` is the right primitive here, a real microtask flush wouldn't prove the old code settles early, so only the name is wrong. `flushAsync` or `nextMacrotask`.

apps/web/src/__tests__/hooks/use-crawler-mutations.test.ts
```diff
@@ -0,0 +1,87 @@
+import { jest, describe, it, expect } from "@jest/globals";
+
+type MutationOptions = {
+  onSuccess?: (...args: unknown[]) => unknown;
+};
+
+/** `useMutation` options captured per procedure name (attachToChatbot, …). */
+const mockCaptured = new Map<string, MutationOptions>();
```
akhileshrangani4, line 8: Nit: worth a `beforeEach(() => mockCaptured.clear())`. Our jest config sets `clearMocks` and `resetMocks`, so the toast spy is fine, but this map is module level and never cleared. Harmless today since every test re-invokes the hook and overwrites its entry, the value is that a capture failure would then show up as an obvious `undefined` instead of silently reusing the previous test's options.

(Related: the Copilot comment about shared mock state is wrong for that same reason, `clearMocks`/`resetMocks` already handle it. You can dismiss it.)

akhileshrangani4 (commented):
@gum1x this is a good one. I went and checked the two library claims against the installed source rather than taking the writeup for them, and both hold.

`@tanstack/query-core` 5.103.1 (we're past the 5.100.8 you cited, main bumped it, same behavior) awaits `options.onSuccess` at `mutation.js:181` before `#dispatch({ type: "success" })` at 184, and `isPending` comes off that state in `mutationObserver.js:137`. On the tRPC side, `createHooksInternal.tsx:87` defaults `mutationSuccessOverride` to `(options) => options.originalFn()` and the wrapper returns its result, and our client is `createTRPCReact<AppRouter>()` with no config, so nothing overrides it. So the mechanism is exactly what you said it was.

I also checked the test actually catches the regression rather than just passing. Pointed jest at the pre-PR hook and the two "holds pending" cases go red with `Expected: false, Received: true` while the two toast ordering cases stay green. That's the right shape.

One real thing to look at, in the first comment below, plus a couple of nits. Nothing blocking.

Answers to your open questions:

- toggle and remove: agree, leave them. They were block bodied at both call sites before and after #458, so that's pre-existing rather than part of this regression. Separate issue is right.
- toast before the await: keeping it first is correct. The toast is confirming the server result, not the refetch. `renameCrawlSource` toasting after is the odd one out, not the model.
- the bulk remove cost: slightly different from how you described it, and in your favour. The old code fired a refetch per iteration too, but `refetch()` defaults to `cancelRefetch: true`, so most were cancelled mid flight. So it isn't N extra requests, it's N that now actually complete and are serialized. Each completed `getCrawlSources` does run `sweepStaleCrawls` plus two queries, so it is real added server work, but at a page size of 10 I'm fine with it. Keeping the buttons disabled for the whole run is worth more.

Last thing, can you merge `main` in? Two chat suites fail on your head SHA and it's nothing to do with you, #497 bumped the `ai` SDK and updated those two test files after your base commit. A sync turns CI green.

PR #509, opened 2026-09-16
fix(auth): add the account columns Better Auth 1.7 validates for

akhileshrangani4:
Fixes the 1.34.13 sign-in outage.

## What broke

`better-auth` 1.6.14 to 1.7.5, shipped in #497, added schema validation to the Drizzle adapter. Three things about it together explain the symptom:

- it is **on by default** (`advanced.database.validateSchema !== false`)
- it runs **on every auth request**, not at startup (`to-auth-endpoints.mjs:41`)
- it **throws** `SchemaMismatchError` when a column it declares is missing, and the result is cached and rethrown on every later call

So a single absent column takes out sign-in, sign-up and session reads at once.

Our `account` table never had `accessTokenExpiresAt`, `refreshTokenExpiresAt` or `scope`. Those are OAuth columns, this app is email and password only, so nothing ever touched them and 1.6 never looked.

Confirmed rather than guessed: running Better Auth's own `findDrizzleSchemaProblems` against our schema and our auth options returns exactly three findings.

```
{ kind: "missing-column", table: "account", column: "accessTokenExpiresAt" }
{ kind: "missing-column", table: "account", column: "refreshTokenExpiresAt" }
{ kind: "missing-column", table: "account", column: "scope" }
```

Worth recording what it was **not**, since the release notes point that way: 1.7.0 did introduce an `issuer` column and an `(issuer, accountId)` key, but 1.7.3 reverted to identifying accounts by `(providerId, accountId)` specifically to restore 1.6 compatibility. Dumping the core table definitions for 1.6.14 and 1.7.5 side by side shows no field differences at all. The migration guide's `issuer` backfill does not apply to us.

## The fix

The three columns, nullable, plus migration 0019. Drizzle selects every declared column by name, so the schema change and the migration have to land together. The release workflow already applies migrations before pushing to the production branch, so that ordering is handled.

I did not take the one-line escape hatch (`advanced: { database: { validateSchema: false } }`). It would silence the check while leaving the schema genuinely out of step with what the library expects, and these columns are needed the moment OAuth is added.

## The test

`auth-schema.test.ts` runs Better Auth's own expected tables against the columns our Drizzle schema declares. No database required.

This is the part I care about more than the fix. Nothing in the pipeline could have caught this:

- the columns are unused by our code, so `check-types` passes
- the test suite has no database, so tests pass
- the check only runs against a live adapter, so `build` passes
- it fails in production, on the first request, or not at all

Verified it actually catches the bug by removing `scope` and watching it fail with `account.scope`, rather than trusting that it would.

`user.additionalFields` moves into its own module so the test can read it without importing the auth instance, which pulls in env, the database and the email client at module load. No behavior change.

## Verification

check-types, lint clean. 1033 tests pass (940 web, 79 ai, 14 db).

Not verified: I could not reproduce an actual sign-in end to end, since my Vercel login cannot see that team's logs and a local run would need the app stood up against the docker database. The evidence here is the library's own validator reporting the mismatch, plus the code path from that validator to every auth endpoint. If you still have the 1.34.13 deployment logs, a `SchemaMismatchError` in them would close the loop.


akhileshrangani4:
Confirmed against production logs. This is the error from the 1.34.13 deployment (`dpl_7S4YDBvdn7RbAuQQgDMmUyz32rRe`), verbatim:

```
Error [BetterAuthError]: Drizzle schema mismatch

  Missing columns
    account.accessTokenExpiresAt
    account.refreshTokenExpiresAt
    account.scope

  help: Run `npx auth generate` to refresh the Drizzle schema, then apply it with your migration tool.
] {
  code: 'SCHEMA_MISMATCH',
  source: 'drizzle'
}
```

**197 occurrences across 16 users on `/api/auth/[...all]`**, from 21:23:43 to 21:51:17 UTC, which is the deployment window up to the rollback. A second group of 15 spans `/api/auth/[...all]` and `/api/trpc/[trpc]`, which is the session read failing and taking authenticated tRPC calls down with it.

The three columns match this PR one for one, so the diagnosis above is confirmed rather than inferred, and the fix is complete.

